PR #5, opened 2020-05-12
BUG: Fix packing/unpacking of prebuilt model build_fn attribute

### LINKED ISSUE #4: A  problem about generating BaggingClassifier with wrapped classifier
Hi, adrian:
Really appreciate your work. i read your some comments in other issues.
However, i have a question. When generating a BaggingClassifier with classifier wrapped from keras:
```
classifier = sklearn_keras_wrap.wrappers.KerasClassifier(build_fn=model, batch_size=64, epochs=1)
classifier.fit(train_x, train_y, batch_size=64, epochs=1)
ensemble_clf = BaggingClassifier(classifier, n_estimators=n_estimators)
ensemble_clf.fit(train_x, train_y)  #error
```` 
An error will occur:  
```
rv = reductor(4)
TypeError: can't pickle _thread.RLock objects
```
It is mentioned many times. Do you have any targeted suggestions for me about modifying bagging.py?
Thank you very much. Look forward your reply.

### PR DESCRIPTION
**Pull request by adriangb**
Fixes #4 

This adds a violation of the Scikit-Learn API: if `build_fn` is a prebuilt model, it is modified before saving. `Model` instances cannot be pickled directly and hence cannot be saved as an attribute in `__init__`, which is what the Scikit-Learn API expects.

### DIFF
--- a/tests/test_wrappers.py
+++ b/tests/test_wrappers.py
@@ -32,7 +32,7 @@ from tensorflow.python.keras.layers import (
     Flatten,
     Input,
 )
-from tensorflow.python.keras.models import Model, Sequential, clone_model
+from tensorflow.python.keras.models import Model, Sequential
 from tensorflow.python.keras.utils.np_utils import to_categorical
 
 from sklearn_keras_wrap import wrappers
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/test_wrappers.py [hunk 1]
```diff
@@ -745,34 +746,27 @@ def test_prebuilt_model(self):
...
-
-            n_classes_ = np.unique(y_train).size
-            # make y the same shape as will be used by .fit
-            if config != "MLPRegressor":
-                y_train = to_categorical(y_train)
-                keras_model = build_fn(
-                    X=x_train, n_classes_=n_classes_, n_outputs_=1
-                )
-            else:
-                keras_model = build_fn(X=x_train, n_outputs_=1)
```
adriangb, line 769: This test was originally added because the failure from using a prebuilt model without training was unintuitive. These changes make the failure a clean error message from Keras, so I see no need to test for it anymore.

tests/test_wrappers.py
```diff
@@ -745,34 +746,27 @@ def test_prebuilt_model(self):
...
+    @pytest.mark.parametrize("config", ["MLPRegressor", "MLPClassifier"])
+    def test_ensemble(self, config):
```
adriangb, line 750: I confirmed that this new test fails on the `master` branch, so I think it reflects the issue reported in #4

adriangb:
This discussion might be relevant: tensorflow/tensorflow#34697

PR #39, opened 2020-08-07
move to poetry and github actions

adriangb:
This PR modernizes the installation and package management by:
1) Switching to `poetry`
2) Switching to GitHub Actions


adriangb:
This is finally starting to look good.

I need to go back and undo some unnecessary code changes.

Also blocked by https://github.com/actions/virtual-environments/issues/1354

tests/test_wrappers.py
```diff
@@ -580,7 +584,12 @@ class FullyCompliantClassifier(wrappers.KerasClassifier):
     def __init__(
         self,
         hidden_dim=HIDDEN_DIM,
-        batch_size=BATCH_SIZE,
+        # Set batch size to a large number (larger than X.shape[0] is the goal)
+        # if batch_size < X.shape[0], results will very
+        # slightly if X is shuffled.
+        # This is only required for this tests and is not really
+        # applicable to real world datasets
+        batch_size=1000,
```
adriangb, line 592: This was picked up from testing with the nightly version of Scikit-Learn

scikeras/__init__.py
```diff
@@ -8,8 +8,7 @@
 # See https://github.com/tensorflow/tensorflow/pull/42097
 # Will be removed whenever the
 # min supported version of tf incorporates the fix
-import tensorflow.python.keras.metrics  # noqa
```
adriangb, line 11: This import is broken in `tf-nightly` as of 20200810. This change fixes that, at least for now.

adriangb:
ccing @stsievert  since I know you are actively developing with this library

adriangb:
I'm happy to say that this was largely successful in speeding up runs, I'm seeing Ubuntu CI runs take as little as 2 minutes, where they were taking ~6 min before. The overall time is still about the same because the slowest runners (it looks like Windows) are still taking ~ 6 min.

It was also great at catching bugs with nightly versions, I got several hits already before I even merged this!

Issue #37, opened 2020-08-06
RFC: time to simplify APIs?


adriangb:
## Background

Currently, this package has many ways to:
- Pass arguments to `build_fn`/`_keras_build_fn`.
- Pass arguments to Keras Models' `fit` or `predict`.

This comes from a combination of supporting the original `tf.keras.wrappers.scikit_learn` interface along with introduction of new ways to do things to improve Scikit-Learn compability.

### Important principles

Since this aims to be a _small_ pure Python package, I think it is important to keep in mind some of Python's guiding principles (cherry picked from [PEP20](https://www.python.org/dev/peps/pep-0020/)):
>  - Explicit is better than implicit.
>  - Simple is better than complex.
> - There should be one-- and preferably only one --obvious way to do it.

For the most part, the APIs mentioned above rely on dynamic parsing of function signatures and filtering of kwargs or attributes by name. This is somewhat "complex" and "implicit" in my opinion.

### Next steps

I _feel_ (and here is where I would appreciate some feedback from others) that it would be good to fully document the requirements that these wrappers have and then narrow down the API to be as simple as possible while still meeting all of those requirements. Off the top of my head:
- Full compatibility with the Scikit-Learn API (this includes hyperparameter optimization type stuff).
- Compatible with Sequential, Functional and Model subclass Keras APIs.
- Ability to use pre-built Keras Models.
- Ability to access wrapper parameters and attributes during building of a Keras Model.
- Ability to pass arguments to the Keras Model's `fit` and `predict` methods.
   - Is there a use case to pass parameters during a call to `fit`, or can we require that they be set from `__init__`? As far as I can tell, only the latter makes sense as far as Scikit-Learn is concerned.

Based on the above reqs (which admittedly could be shortsighted), it seems reasonable to me to take the following action:
- Remove `**kwargs` from `fit` and `predict` and require that these parameters be set via `__init__`.
- Remove `**kwargs` from `__init__` and instead hardcode Keras Model `fit` and `predict` parameters (maybe also `compile` parameters?) as proposed in #30.

A more extreme step would be to remove the `build_fn` argument and force users to always use the subclassing interface since it is technically even possible to return a pre-built model from `_keras_build_fn` via a closure or other methods. This would greatly simplify the API but I worry that it would be an inconvenience for users (even if it is just a couple more lines of code).

All in all I hope to reduce codebase complexity and simplify documentation. Any comments are welcome.

adriangb:
[...]

PR #47, opened 2020-08-12
REF: Hardcode Keras params

adriangb:
This implements hardcoding of common Keras parameters in `BaseWrapper.__init__` as discussed in #37.

The parameters chosen come from `Model.compile`, `Model.predict` and `Model.fit`. Parameters were not hardcoded if:
- They don't make sense for Scikit-Learn/Numpy arrays (for example, `steps_per_epoch`).
- Should be fed to `fit` directly because they depend on the data (ex: `sample_weights`, `class_weights`). I think `validation_data` also falls into this category, but I am not sure if this makes much sense to implement since AFAIK this is not used anywhere else in Scikit-Learn. I am going to leave this for another PR where I'll also deprectate `fit`'s `**kwargs`.

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -29,6 +29,7 @@ from tensorflow.python.keras.utils.generic_utils import (
 
 from ._utils import LabelDimensionTransformer
 from ._utils import TFRandomState
+from ._utils import _get_default_args
 from ._utils import get_metric_full_name
 from ._utils import make_model_picklable
 
@@ -124,8 +125,11 @@ class BaseWrapper(BaseEstimator):
 
         self.build_fn = build_fn
 
-        if sk_params:
+        if inspect.isfunction(build_fn):
+            kwargs = _get_default_args(build_fn)
+            sk_params = {**kwargs, **sk_params}
 
+        if sk_params:
             # for backwards compatibility
 
             # the sklearn API requires that all __init__ parameters be saved
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -103,14 +103,6 @@ class BaseWrapper(BaseEstimator):
         # Unpack kwargs
         vars(self).update(**kwargs)
 
-        # check that all __init__ parameters were assigned (as per sklearn API)
-        params = self.get_params(deep=False)
-        for key in params.keys():
-            try:
-                getattr(self, key)
-            except AttributeError:
-                raise RuntimeError("Unasigned input parameter: {}".format(key))
-
     @property
     def __name__(self):
         return self.__class__.__name__
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -87,8 +87,7 @@ class BaseWrapper(BaseEstimator):
 
         # Get defaults from `build_fn`
         if inspect.isfunction(build_fn):
-            kwargs = _get_default_args(build_fn)
-            sk_params = {**kwargs, **sk_params}
+            kwargs.upadate(_get_default_args(build_fn))
 
         # Parse hardcoded params
         self.random_state = random_state
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -90,14 +90,6 @@ class BaseWrapper(BaseEstimator):
             kwargs = _get_default_args(build_fn)
             sk_params = {**kwargs, **sk_params}
 
-        # check that all __init__ parameters were assigned (as per sklearn API)
-        params = self.get_params(deep=False)
-        for key in params.keys():
-            try:
-                getattr(self, key)
-            except AttributeError:
-                raise RuntimeError("Unasigned input parameter: {}".format(key))
-
         # Parse hardcoded params
         self.random_state = random_state
         self.optimizer = optimizer
@@ -117,6 +109,14 @@ class BaseWrapper(BaseEstimator):
 
         # Restore parameters already set before this __init__
         vars(self).update(**existing_params)
+
+        # check that all __init__ parameters were assigned (as per sklearn API)
+        params = self.get_params(deep=False)
+        for key in params.keys():
+            try:
+                getattr(self, key)
+            except AttributeError:
+                raise RuntimeError("Unasigned input parameter: {}".format(key))
 
     @property
     def __name__(self):
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -20,7 +20,6 @@ from sklearn.utils.validation import check_array
 from sklearn.utils.validation import check_X_y
 from tensorflow.keras import backend as k_backend
 from tensorflow.keras.models import Model
-from tensorflow.keras.models import Sequential
 from tensorflow.python.keras.losses import is_categorical_crossentropy
 from tensorflow.python.keras.utils.generic_utils import has_arg
 from tensorflow.python.keras.utils.generic_utils import (
@@ -32,15 +31,6 @@ from ._utils import TFRandomState
 from ._utils import _get_default_args
 from ._utils import get_metric_full_name
 from ._utils import make_model_picklable
-
-
-# known keras function names that will be added to _legal_params_fns if they
-# exist in the generated model
-KNOWN_KERAS_FN_NAMES = (
-    "fit",
-    "evaluate",
-    "predict",
-)
 
 
 class BaseWrapper(BaseEstimator):
@@ -58,6 +48,7 @@ class BaseWrapper(BaseEstimator):
             Set the Tensorflow random number generators to a
             reproducible deterministic state using this seed.
             Pass an int for reproducible results across multiple function calls.
+        For all other parameters see tf.keras.Model documentation.
     """
 
     is_fitted_ = False
@@ -131,7 +122,7 @@ class BaseWrapper(BaseEstimator):
         self.initial_epoch = initial_epoch
         # Unpack kwargs
         vars(self).update(**kwargs)
-        # Restrore parameters already set before this __init__
+        # Restore parameters already set before this __init__
         vars(self).update(**existing_params)
 
     @property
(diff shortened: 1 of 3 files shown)


tests/test_wrappers.py
```diff
@@ -1224,24 +1223,6 @@ def testpostprocess_y(self):
...
-        class ClassBuildFnClf(wrappers.KerasClassifier):
-            def __init__(self, input_param):
-                # does not set input_param
-                super().__init__()
-
-            def _keras_build_fn(self, hidden_dim):
-                return build_fn_clf(hidden_dim)
-
-        with pytest.raises(RuntimeError):
-            ClassBuildFnClf(input_param=10)
```
adriangb, line 1242: Removing this check and test. It is really a redundant feature that no other estimators have. It was much more helpful previously when the interface was less defined.

tests/test_wrappers.py
```diff
@@ -417,9 +417,12 @@ class ClassWithCallback(wrappers.KerasClassifier):
...
     """
 
-    def __init__(self, **sk_params):
-        self.callbacks = [SentinalCallback()]
-        super().__init__(**sk_params)
+    def __init__(self, callbacks=None, hidden_dim=(100,)):
+        if callbacks is None:
+            self.callbacks = [SentinalCallback()]
+        else:
+            self.callbacks = callbacks
```
adriangb, line 424: I think this brings up an interesting question that is not a blocker for this PR but will be interesting for param routing: a quite common pattern in scikit-learn estimators is to parse/modify an `__init__` param in `fit`, ex:
```python
def fit(X, y, ...):
    self.random_state_ = np.random.RandomState(seed=self.random_state)
```

Where downstream the estimator will only use the `self.random_state_` param and not `self.random_state`.

I think it would be nice to allow users to do something like:

```python3
class MyClassifier(KerasClassifier):

    def hook_params(self):   # in base class this would just be a pass
         if self.batch_size is None:
              self.batch_size_ = 32
         else:
              self.batch_size_ = self.batch_size
```

But the trick will then be routing that parameter so that `callbacks_` always overrides `callbacks`. Maybe if two parameters match except for `_`'s at the end, the one with the most `_` takes precedence?  So that:

```python3
def build_fn(batch_size=32):
    print(f"batch_size={batch_size}")
    ...

class MyClassifier(KerasClassifier):

    def hook_params(self):   # in base class this would just be a pass
         if self.batch_size is None:
              self.batch_size_ = 32
         else:
              self.batch_size_ = self.batch_size
         self.batch_size__ = 64

estimator = MyClassifier()
estimator.fit(X, y)  # always prints `batch_size=64`
```

scikeras/wrappers.py
```diff
@@ -676,11 +638,11 @@ def _filter_params(self, fn, params_to_check=None):
...
-        parameters = super()._get_param_names()
-        # add kwargs/sk_params if the user gave those as input
-        if self._sk_params:
-            return sorted(parameters + self._sk_params)
-        return parameters
+        return (
+            k
+            for k in self.__dict__
+            if not k.endswith("_") and not k.startswith("_")
+        )
```
adriangb, line 639: This is borrowed from SciKeras. For where we are headed, I think this makes more sense than storing `_sk_params`. It'll be up to users to make sure they don't set any params without `_` that they don't want to show up in `get_params`.

scikeras/wrappers.py
```diff
@@ -51,94 +41,67 @@ class BaseWrapper(BaseEstimator):
...
+        self.batch_size = batch_size
+        self.verbose = verbose
+        self.callbacks = callbacks
+        self.epochs = epochs
+        self.validation_split = validation_split
+        self.shuffle = shuffle
+        self.initial_epoch = initial_epoch
+
+        # Unpack kwargs
+        vars(self).update(**kwargs)
```
adriangb, line 104: Also borrowed from scikeras. So much cleaner with the new `_get_param_names` implementation.

adriangb:
tagging @stsievert in case you want to take a look

tests/test_wrappers.py
stsievert: Fyi, sklearn.utils has a "check_random_state" function. I don't recall estimators ever setting a "random_state_" attribute.

adriangb: Yeah, I was just looking for a clear example, I didn't check that it's a real one.

adriangb: I updated it to a real-world example from Scikit-Learn.

stsievert: Thanks for the update; I also do that in my estimators.

adriangb: Again, not for this PR, but certainly a required feature for parameter routing. I'm going to copy this discussion to #50 .

scikeras/wrappers.py
adriangb: I added `k.startswith("_")` to accomodate parameters like `self._random_state`. So the rule will be that all model paramers (i.e. the `__init__` ones) cannot start/end with `_`, at least for now. We'll have to see how that plays in with my overwriting proposal in https://github.com/adriangb/scikeras/pull/47#discussion_r469709646

adriangb:
@stsievert do you think we are good to go here or is there anything else you see that should be changed?

scikeras/wrappers.py
```diff
@@ -51,102 +41,67 @@ class BaseWrapper(BaseEstimator):
...
+        random_state=None,
+        optimizer="rmsprop",
+        loss=None,
+        metrics=None,
+        run_eagerly=None,
+        batch_size=None,
+        verbose=1,
+        steps=None,
+        callbacks=None,
+        epochs=1,
```
stsievert, line 74: Maybe rename to `max_epochs` for the `fit` implementation. I think `partial_fit` should always have `epochs=1`.

scikeras/wrappers.py
```diff
@@ -51,102 +41,67 @@ class BaseWrapper(BaseEstimator):
...
+        validation_split=0.0,
+        shuffle=True,
+        initial_epoch=0,
```
stsievert, line 77: This only makes sense if a pretrained model is passed to `build_fn`. I'd be okay removing support for this, especially if `max_epochs` is supported.

scikeras/wrappers.py
```diff
@@ -51,102 +41,67 @@ class BaseWrapper(BaseEstimator):
...
-    def __init__(self, build_fn=None, **sk_params):
+    def __init__(
+        self,
+        build_fn=None,
+        *,
+        random_state=None,
+        optimizer="rmsprop",
+        loss=None,
+        metrics=None,
+        run_eagerly=None,
```
stsievert, line 69: Does SciKeras have tests for `run_eagerly in [True, False]`? It seems `run_eagerly` will affect the serialization.

scikeras/wrappers.py
```diff
@@ -51,102 +41,67 @@ class BaseWrapper(BaseEstimator):
...
     }
 
-    def __init__(self, build_fn=None, **sk_params):
+    def __init__(
+        self,
+        build_fn=None,
+        *,
+        random_state=None,
+        optimizer="rmsprop",
+        loss=None,
```
stsievert, line 67: Where are `optimizer` and `loss` used? Are they passed to the `compile` method?

scikeras/wrappers.py
adriangb: No, we do not. Maybe I'll just remove that parameter for now, I don't want to go down that rabbit hole in this PR...

scikeras/wrappers.py
adriangb: As it stands right now, it would be up to the user to declare their `build_fn` to use them. I.e.:

```python3
def build_fn(optimizer=..., loss=...):
    model = Model()
    ...
    model.compile(loss=loss, optimizer=optimizer)
```

adriangb:
> There might be some missing parameters

The reasons they are not being included are:
* `steps_per_epoch`: the docs say `This argument is not supported with array inputs` so I think it only makes sense for `tf.data`, which we do not support here.
* `sample_weight`: this is/will still be handled in `fit` like Scikit-Learn does.
* `class_weight`: I think this one requires a bit of massaging to make Scikit-Learn's interpretation of this ([ref](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html)) match up with Keras' ([ref](https://www.tensorflow.org/api_docs/python/tf/keras/Model)). I think Keras maps "integer class indices" to weights while Scikit-Learn maps "class labels" to weights. I'd like to leave this for a separate PR.

Regarding `max_epochs`/`epochs`.
I think what Keras does is that it runs for `epochs-initial_epoch` epochs. Are you proposing that instead we just have a single parameter `max_epochs` and always start at epoch 0?
And separately, you are suggesting that `partial_fit` override whatever `max_epochs` was set to and pass `epochs=1` to `model_.fit`?

stsievert:
> And separately, you are suggesting that partial_fit override whatever max_epochs was set to and pass epochs=1 to model_.fit?

Yes. Scikit-Learn's SGDClassifier/SGDRegressor/MLPClassifier do the same thing. Here's the docstring for [MLPClassifier.partial_fit][1]: "Update the model with a single iteration over the given data."

> I think what Keras does is that it runs for `epochs - initial_epoch` epochs. Are you proposing that instead we just have a single parameter max_epochs and always start at epoch 0?

I think the difference is semantic. I don't think it's super important. `max_epochs` provides a little more leeway to perform some early stopping. Sticking with the Keras API isn't a bad choice I think.

I mention it because Scikit-Learn calls this parameter `max_iter`. In the documentation they always mention `max_iter` is "the maximum number of passes over the training data (aka epochs)." 

[1]:https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html#sklearn.neural_network.MLPClassifier.partial_fit

adriangb:
Hmm okay. I think what I'd like to do is keep this PR smallish and leave the following parameters for a future PR (that can also contain any logic related to their handling, like what you mention about `fit` and epochs).
* `class_weight`
* `epochs`/`initial_epoch`/`max_epochs`
So I'll remove `epochs` from here.

PR #56, opened 2020-08-15
REF: Split up tests into modules & remove redundant testing

### LINKED ISSUE #43: TST: Inconsistent failing for TestSampleWeights
TestSampleWeights keeps failing randomly both on master and other branches. Overall, the test is not very well designed, so I think it needs some re-work. I'm going to disable it for now as part of #39 , but I'm opening this issue to re-work or delete it in the future.

adriangb:
* Split up tests into multiple modules
* Implement scikit-learn estimator checks on non-subclassed estimators
* Remove redundant tests and test parametrizations.
* Fix #43 by implementing a much better test.
* If pytest-xdist fails on CI, still fail the run but also try to run tests without xdist to get clean debugging info.


tests/test_scikit_learn_checks.py
```diff
@@ -0,0 +1,79 @@
...
+            # This is only required for this tests and is not really
+            # applicable to real world datasets
+            batch_size=1000,
+            optimizer="adam",
+            loss=KerasRegressor.root_mean_squared_error,
+        ),
+    ],
+    ids=["KerasClassifier", "KerasRegressor"],
+)
+def test_fully_compliant_estimators(estimator, check):
```
adriangb, line 46: @stsievert I know we went back and forth on checking that non-subclassed models can pass all tests. Here is proof that they do. As you suggested, the only test they fail is `check_no_attributes_set_in_init`. I added a separate tests specifically for that below.

PR #59, opened 2020-08-15
MAINT: remove input dtype checks

stsievert:
It removes checks on the input data. What if `float32` data is passed? That's common in ML.

TODO:

- [x] Can we use OrdinalEncoder instead of LabelEncoder to avoid one more dtype conversion?
- [x] Test for KerasRegressor handling of integer `X` and/or `y`
- [x] Test for KerasClassifier handling of object/string `y`


adriangb:
Thank you for this! I agree that we should allow any data type. Will we need to handle data conversions to make sure `X` and `y` have the same type? Some tests are failing now because of that. I realize that in general users should not be feeding `X` and `y` of different types, but since most Scikit-Learn estimators do not have any problems with that I believe that here we should figure out how to allow that as well.

stsievert:
> Will we need to handle data conversions to make sure X and y have the same type?

Do types always have to be the same for the input and output? It'd probably be fair to use [np.result_type][1] for typecasting.

[1]:https://numpy.org/doc/stable/reference/generated/numpy.result_type.html

Does the input/output type matching change with `run_eagerly`?

adriangb:
> Do types always have to be the same for the input and output? It'd probably be fair to use np.promote_types for typecasting.
Does the input/output type matching change with run_eagerly?

I am not 100% sure but I think the answers are yes and no respectively.

adriangb:
@stsievert I implemented an initial prototype for a type conversion-deconversion system. It could probably benefit from some more thought before being implemented.

stsievert:
Why do we need an implementation for type-casting? The error message is clear and easy to resolve:

> TypeError: Input 'y' of 'Sub' Op has type float32 that does not match type int64 of argument 'x'.

Plus, this test passes on 4f0329d :

``` python
@pytest.mark.parametrize("X_dtype", ["float32", "float64"])
@pytest.mark.parametrize("y_dtype", ["int64", "int32", "uint8", "uint16"])
@pytest.mark.parametrize("run_eagerly", [True, False])
def test_classifier_handles_types(X_dtype, y_dtype, run_eagerly):
    clf = KerasClassifier(build_fn=dynamic_classifier, run_eagerly=run_eagerly)
    n, d = 100, 20
    n_classes = 10
    X = np.random.uniform(size=(n, d)).astype(X_dtype)
    y = np.random.choice(n_classes, size=n).astype(y_dtype)
    clf.fit(X, y)
    assert clf.score(X, y) >= 0
```

It looks like `y` is being type-cast to a float with `LabelEncoder`.`self.model_.fit` is being called with `(X.dtype, y.dtype) == ("float32", "float64")`  when the inputs dtypes are `(X.dtype, y.dtype) == ("float32", "uint8")`.

adriangb:
Thank you for the input. I went back and did some more testing and determined a couple of things:
- Scikit-Learn only really expects the output dtype for `y` to match it's input dtype.
- The TF failures were due to our custom R2 loss function, which could not handle different dtypes. Because of how TF builds the graphs, the error doesn't show up in `wrappers.py` and instead shows up inside TF, which confused me and I would imagine would confuse a user as well.

With this knowledge, I was able to remove this type casting system and have only a single cast back to the input dtype for `y` in `KerasClassifier.postprocess_y`. Now all tests are passing, including the one you propose in https://github.com/adriangb/scikeras/pull/59#issuecomment-674540460. I still want to add a couple more checks, so I'm not pushing yet, but I'm posting just to make sure you don't spend more time on this than needed.

adriangb:
I'm seeing some failures on Windows now, it looks like we're running into https://github.com/tensorflow/probability/issues/886

scikeras/wrappers.py
```diff
@@ -250,6 +255,23 @@ def _fit_keras_model(self, X, y, sample_weight, warm_start, **kwargs):
...
+            else:
+                X = X.astype(np.int64) if X.dtype == np.int32 else X
+            if not isinstance(y, np.ndarray):  # list, tuple, etc.
+                y = [
+                    y_.astype(np.int64) if y_.dtype == np.int32 else y_
+                    for y_ in y
+                ]
+            else:
+                y = y.astype(np.int64) if y.dtype == np.int32 else y
+
```
adriangb, line 274: The windows failures are now passing with this hack... I think this is an actual bug in TF that they should fix, but I guess this will do for now...

adriangb:
@stsievert I think this is ready for you to take another look

scikeras/wrappers.py
```diff
@@ -834,6 +855,9 @@ def postprocess_y(self, y):
 
         y = np.squeeze(np.column_stack(class_predictions))
 
+        # type cast back to input dtype
+        y = y.astype(self.input_dtype_, copy=False)
```
stsievert, line 859: :+1: This will preserve the input/output dtype of `y`, which might be important for pipelines.

I've added the flag `copy=False`. There's no need to copy if it's of the same dtype (which is what `astype` does by default).

tests/test_input_outputs.py
```diff
@@ -251,3 +254,71 @@ def test_BaseWrapper_postprocess_y():
...
+            assert y.dtype == input_dtypes[1]
+            assert sample_weight.dtype == input_dtypes[2]
+            return super()._fit_keras_model(
+                X, y, sample_weight, warm_start, **kwargs
+            )
+
+    clf = KerasRegressor(build_fn=dynamic_regressor, run_eagerly=run_eagerly)
+    clf.fit(X, y, sample_weight=sample_weight)
+    return_dtype = clf.predict(X).dtype
+    assert return_dtype.kind in ("f", "c")
```
stsievert, line 324: Don't we want `assert return_dtype == y.dtype`? Where does complex data enter here?

scikeras/wrappers.py
```diff
@@ -250,6 +255,23 @@ def _fit_keras_model(self, X, y, sample_weight, warm_start, **kwargs):
         # order implies kwargs overwrites fit_args
         fit_args = {**fit_args, **kwargs}
 
+        if OS_IS_WINDOWS:
+            # see tensorflow/probability#886
+            if not isinstance(X, np.ndarray):  # list, tuple, etc.
```
stsievert, line 260: I might make a small helper function to remove some for-loops/if-staments:

``` python
def _windows_upcast_ints(x: np.ndarray) -> np.ndarray:
    return x.astype("int64") if x.dtype == np.int32 else x

if OS_IS_WINDOWS:
    X = _windows_upcast_int(X) if isinstance(X, np.ndarray) else [_windows_upcast_int(x) for x in X]
    y = _windows_upcast_int(y) if isinstance(y, np.ndarray) else [_windows_upcast_int(yi) for yi in y]
```

adriangb: Yeah that would help, please feel free to implement!

scikeras/wrappers.py
```diff
@@ -688,7 +708,7 @@ def preprocess_y(y):
             encoder = LabelEncoder()
```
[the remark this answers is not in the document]
stsievert: If not, I'd add `copy=False` to all of the `astype` calls below.

tests/test_input_outputs.py
adriangb: It's expected that if your input y to the regressor was `int32` you will _not_ get `int32` back and instead you'll get a float. I'm 98% sure there would never be a complex float but I included it just in case. If you're 100% sure then by all means make it just `return_dtype.kind == "f"`

scikeras/wrappers.py
adriangb: Good catch. Small optimizations add up!

scikeras/wrappers.py
```diff
@@ -760,6 +780,7 @@ def preprocess_y(y):
             "keras_expected_n_ouputs_": keras_expected_n_ouputs_,
             "n_classes_": n_classes_,
             "cls_type_": cls_type_,
+            "input_dtype_": input_dtype_,
```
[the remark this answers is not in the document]
adriangb: Noted and agreed.

scikeras/wrappers.py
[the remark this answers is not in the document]
adriangb: We *might* be able to use `OrdinalEncoder`, I think the output might always be at least 2D but reshaping is certainly cheaper than _always_ having to cast.

scikeras/wrappers.py
```diff
@@ -33,6 +34,9 @@
 from ._utils import make_model_picklable
 
 
+OS_IS_WINDOWS = os.name == "nt"  # see tensorflow/probability#886
```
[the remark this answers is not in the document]
adriangb: https://github.com/adriangb/scikeras/runs/990863534?check_suite_focus=true

scikeras/wrappers.py
adriangb: Unfortunately, I did not see one. It might be as easy as:

```python3
import numpy as np
from tensorflow.python.framework.constant_op import convert_to_eager_tensor


convert_to_eager_tensor(np.array([1], dtype=np.int32)
```

tests/test_input_outputs.py
adriangb: I just realized that this is just our _test_ and that we do not use complex floats here so... totally safe to remove `"c"`. I also think complex floats are not supported in general (see [here](https://stackoverflow.com/questions/11999147/is-it-possible-to-use-complex-numbers-as-target-labels-in-scikit-learn)).

stsievert: I've made this test stronger in d274a1c7884eac684b873491289c37aba741fd0f, and an accidental casting to `float64`. This relies on the fact that regression problems require floating point outputs.

tests/test_input_outputs.py
```diff
@@ -251,3 +254,70 @@ def test_BaseWrapper_postprocess_y():
...
+    clf.fit(X, y, sample_weight=sample_weight)
+    assert clf.score(X, y) >= 0
+    assert clf.predict(X).dtype == y.dtype
+
+
+@pytest.mark.parametrize(
+    "X_dtype", ["float32", "float64", "int64", "int32", "uint8", "uint16"]
+)
+@pytest.mark.parametrize("y_dtype", ["float32", "float64"])
+@pytest.mark.parametrize("s_w_dtype", ["float32", "float64"])
```
adriangb, line 298: Can we keep tests for other int dtypes?

tests/test_input_outputs.py
```diff
@@ -251,3 +254,70 @@ def test_BaseWrapper_postprocess_y():
     np.testing.assert_equal(y_postprocessed, y_array)
     extra_args = BaseWrapper.postprocess_y(y_array)[1]
     assert len(extra_args) == 0
+
+
+@pytest.mark.parametrize(
+    "X_dtype", ["float32", "float64", "int64", "int32", "uint8", "uint16"]
+)
+@pytest.mark.parametrize("y_dtype", ["int64", "int32", "uint8", "uint16"])
```
adriangb, line 262: Should we have tests for string/object dtype?

scikeras/wrappers.py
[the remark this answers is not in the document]
adriangb: Actually, the dtype here does not matter. Even if the user passed in `float32` (which doesn't make much sense for a categorical variable), the classes are going to be encoded into integers, so `int64` makes sense. I removed that check in the test, the test was _too_ introspective I think.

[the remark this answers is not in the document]
adriangb: What we could do which maybe would be _marginally_ better would be to use the number of classes to determine how large of a number we need to hold and then pick an output dtype to use with `OrdinalEncoder` based on that. But that sounds complicated.

tests/test_input_outputs.py
stsievert: Maybe. Right now I'm getting failures when I cast with `y.astype(str)`. It looks like there's some errors in `LabelEncoder` because it assumes the array is sparse.

tests/test_input_outputs.py
stsievert: How is an integer output array `y` relevant for a regression problem? I think floating point values are far more relevant.

adriangb: Output Is not relevant. Outputs should always be float. But it should accept int inputs.

adriangb:
I think this is looking great! Thank you @stsievert  for all of your work.

scikeras/wrappers.py
```diff
@@ -314,14 +335,36 @@ def _validate_data(self, X, y=None, reset=True):
...
                 y,
                 allow_nd=True,  # allow X to have more than 2 dimensions
                 multi_output=True,  # allow y to be 2D
+                dtype=None,
+            )
+            y = check_array(
+                y, ensure_2d=False, allow_nd=False, dtype=get_dtype(y)
             )
-        X = check_array(X, allow_nd=True, dtype=["float64", "int"])
+        X = check_array(X, allow_nd=True, dtype=get_dtype(X))
```
stsievert, line 367: What's the function `get_dtype` for? Why not pass `dtype="numeric"` to `check_array`? At first glance it looks like `check_array` does the same thing as `get_dtype`:

> **dtype: <i>string, type, list of types or None (default=”numeric”)</i>**
> Data type of result. If None, the dtype of the input is preserved. If “numeric”, dtype is preserved unless array.dtype is object. If dtype is a list of types, conversion on the first type is only performed if the dtype of the input is not in the list.
>
> —[sklearn.utils.check_array docs][1]

It looks like the outputs of `check_array(X, dtype=get_dtype(X))` and `check_array(X, dtype="numeric")` only differ when `X` has an object dtype (in which case they output objects of  `backend.floatx()` and `float64` respectively). Is that accurate?

[1]:https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_array.html

scikeras/wrappers.py
```diff
@@ -314,14 +335,36 @@ def _validate_data(self, X, y=None, reset=True):
         out : {ndarray, sparse matrix} or tuple of these
             The validated input. A tuple is returned if `y` is not None.
         """
+        # default to TFs backend float type
+        # instead of always float64 (sklearns default)
+        tf_backend_dtype = np.dtype(tf.keras.backend.floatx())
+
+        def get_dtype(arr) -> np.dtype:
```
stsievert, line 342: I think there's a typo in this function: it looks like a `np.ndarray` with `arr.dtype.kind == "O"` will needlessly create another array. Maybe this function instead?

``` python
def get_dtype(arr):
    if isinstance(arr, np.ndarray):
        if arr.dtype.kind != "O":
            return arry.dtype
        else:
            return output_dtype

    # arr is not an ndarray
    arr_dtype = np.asarray(arr).dtype
    if arr_dtype.kind != "O":
        return arr_dtype
    return output_dtype
```

scikeras/wrappers.py
```diff
@@ -998,10 +1060,22 @@ def r_squared(y_true, y_pred):
...
         """
-        ss_res = k_backend.sum(k_backend.square(y_true - y_pred), axis=0)
-        ss_tot = k_backend.sum(
-            k_backend.square(y_true - k_backend.mean(y_true, axis=0)), axis=0
+        # Ensure input dytpes match
+        dtype_y_true = np.dtype(y_true.dtype.as_numpy_dtype())
+        dtype_y_pred = np.dtype(y_pred.dtype.as_numpy_dtype())
+        dest_dtype = np.promote_types(dtype_y_pred, dtype_y_true)
+        y_true = tf.cast(y_true, dtype=dest_dtype)
+        y_pred = tf.cast(y_pred, dtype=dest_dtype)
```
stsievert, line 1068: Why is this batch of type-casting necessary? Does the backend calculation of R^2 require it?

scikeras/wrappers.py
adriangb: No, that's on purpose. This is for when we get an iterable of numpy arrays, for example a list of numpy arrays. I tried doing `arr[0].dtype` but that would then fail some tests from the Scikit-Learn checks that specifically check that you do not try to index the inputs before converting them to an array.

scikeras/wrappers.py
adriangb: It requires `y_true` and `y_pred` to match dtypes. We could just cast to `float32` (since I think `y_pred` will always be `float32`) instead of trying to figure out the destination dtype using `np.promote_types`.

scikeras/wrappers.py
adriangb: I misread your comment. You're right. There is a needless conversion. Let me make another PR to fix that.

scikeras/wrappers.py
adriangb: Exactly. The outcome we want is:
- object input: float32 output
- numeric input: same type as output

PR #63, opened 2020-08-17
MAINT: fix needless ndarray creation

adriangb:
As per:
- https://github.com/adriangb/scikeras/pull/59#discussion_r471756149
- https://github.com/adriangb/scikeras/pull/59#discussion_r471768197

@stsievert let me know if there's anything else you catch from #59 . I should have waited for you to merge, sorry about that!

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -344,13 +344,15 @@ class BaseWrapper(BaseEstimator):
             dtype for an input `arr`.
             """
             output_dtype = tf_backend_dtype
-            if isinstance(arr, np.ndarray) and arr.dtype.kind != "O":
-                output_dtype = arr.dtype
-            else:
-                arr_dtype = np.array(arr).dtype
-                if arr_dtype.kind != "O":
-                    # list of arrays with equal dtype
-                    output_dtype = arr_dtype
+            if isinstance(arr, np.ndarray):
+                if arr.dtype.kind != "O":
+                    return arr.dtype
+                else:
+                    return output_dtype
+            # arr is not an ndarray
+            arr_dtype = np.asarray(arr).dtype
+            if arr_dtype.kind != "O":
+                return arr_dtype
             return output_dtype
 
         if y is not None:
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -1063,11 +1063,8 @@ class KerasRegressor(BaseWrapper):
         advisable to use the same loss/metric when optimizing the model.
         """
         # Ensure input dytpes match
-        dtype_y_true = np.dtype(y_true.dtype.as_numpy_dtype())
-        dtype_y_pred = np.dtype(y_pred.dtype.as_numpy_dtype())
-        dest_dtype = np.promote_types(dtype_y_pred, dtype_y_true)
-        y_true = tf.cast(y_true, dtype=dest_dtype)
-        y_pred = tf.cast(y_pred, dtype=dest_dtype)
+        # y_pred will always be float32 so we cast y_true to float32
+        y_true = tf.cast(y_true, dtype=y_pred.dtype)
         # Calculate R^2
         ss_res = tf.math.reduce_sum(
             tf.math.squared_difference(y_true, y_pred), axis=0
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -335,25 +335,16 @@ class BaseWrapper(BaseEstimator):
         out : {ndarray, sparse matrix} or tuple of these
             The validated input. A tuple is returned if `y` is not None.
         """
-        # default to TFs backend float type
-        # instead of always float64 (sklearns default)
-        tf_backend_dtype = np.dtype(tf.keras.backend.floatx())
 
-        def get_dtype(arr) -> np.dtype:
-            """Helper function to determine the correct
-            dtype for an input `arr`.
-            """
-            output_dtype = tf_backend_dtype
-            if isinstance(arr, np.ndarray):
-                if arr.dtype.kind != "O":
-                    return arr.dtype
-                else:
-                    return output_dtype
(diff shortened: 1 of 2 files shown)


stsievert:
Why not this implementation of `get_dtype`? If I'm correct, it the same features expressed in https://github.com/adriangb/scikeras/pull/59#discussion_r471770832

``` python
dtype = None if isinstance(X, np.ndarray) and X.dtype.kind != "O" else tf.keras.backend.floatx()
X = check_array(X, allow_nd=True, dtype=dtype)
```

adriangb:
> Why not this implementation of `get_dtype`? If I'm correct, it the same features expressed in [#59 (comment)](https://github.com/adriangb/scikeras/pull/59#discussion_r471770832)
> 
> ```python
> dtype = None if isinstance(X, np.ndarray) and X.dtype.kind != "O" else tf.keras.backend.floatx()
> X = check_array(X, allow_nd=True, dtype=dtype)
> ```

I *think* this would not work for `X = [np.array([0], dtype=np.int32)`? `dtype` would be `tf.keras.backend.floatx()` when it should be `int32`

stsievert:
Why not this implementation?

``` python
def _check_array_dtype(arr):
    if not isinstance(arr, np.ndarray):
        return _check_array_dtype(np.asarray(arr))
    elif arr.dtype.kind != "O":
        return None  # check_array won't do any casting with dtype=None
    else:
        return tf.keras.backend.floatx()
```

adriangb:
> Why not this implementation?

That is a bit cleaner, the single recursion helps. Thank you.

scikeras/wrappers.py
```diff
@@ -987,12 +980,6 @@ class KerasRegressor(BaseWrapper):
         }
     )
 
-    def _validate_data(self, X, y=None, reset=True):
-        """Convert y to float, regressors cannot accept int."""
-        if y is not None:
-            y = check_array(y, ensure_2d=False)
-        return super()._validate_data(X=X, y=y, reset=reset)
-
```
adriangb, line 995: This can actually be removed now, thanks to the changes in `BaseWrapper._validate_data`!

scikeras/wrappers.py
```diff
@@ -1061,11 +1048,8 @@ def r_squared(y_true, y_pred):
         advisable to use the same loss/metric when optimizing the model.
         """
         # Ensure input dytpes match
-        dtype_y_true = np.dtype(y_true.dtype.as_numpy_dtype())
-        dtype_y_pred = np.dtype(y_pred.dtype.as_numpy_dtype())
-        dest_dtype = np.promote_types(dtype_y_pred, dtype_y_true)
-        y_true = tf.cast(y_true, dtype=dest_dtype)
-        y_pred = tf.cast(y_pred, dtype=dest_dtype)
+        # y_pred will always be float32 so we cast y_true to float32
+        y_true = tf.cast(y_true, dtype=y_pred.dtype)
```
adriangb, line 1052: Always casting to `float32` works. I think that trying to match the dtypes here was overkill and probably less efficient.

tests/test_input_outputs.py
```diff
@@ -296,13 +296,16 @@ def _fit_keras_model(self, X, y, sample_weight, warm_start, **kwargs):
...
 )
 @pytest.mark.parametrize(
-    "y_dtype", ["float32", "float64", "int64", "int32", "uint8", "uint16"]
+    "y_dtype",
+    ["float32", "float64", "int64", "int32", "uint8", "uint16", "object"],
 )
 @pytest.mark.parametrize(
-    "s_w_dtype", ["float32", "float64", "int64", "int32", "uint8", "uint16"]
+    "s_w_dtype",
+    ["float32", "float64", "int64", "int32", "uint8", "uint16", "object"],
```
adriangb, line 308: These tests now take a _long_ time with all of these dtypes. I think we need to split this up into:
1) Test a single dtype at a time for all inputs (i.e. `y`, `X` and `sw` all have `float32`, etc.).
2) Test that mixed dtypes work (we can do 1 float and 1 int for each input, giving 6 tests)

I'm not sure where/how we need to be testing for `run_eagerly`, but if we can reduce that parametrization as well I think that would be nice.

stsievert: Yeah, I've noticed that these tests take a long time too.

I think `run_eagerly` should be tested, certainly with different input types. `run_eagerly` controls if the program is compiled in Tensorflow ahead of time, and if TF can do dynamic type inference.

adriangb: Do you think it needs to be tested against both of the tests proposed above, or can we pick the higher risk one?

stsievert: I think `run_eagerly` only needs to be tested on (2) above. But you could include one set of parameters with the same dtype just make sure it passes:

``` python
@pytest.mark.parametrize("y_dtype", ["float32", "float64", "uint8", "int16"])
@pytest.mark.parametrize("run_eagerly", [True, False])
def test_mixed_dtypes(y_dtype, run_eagerly):
    n, d = 100, 10
    n_classes = 4
    X = np.random.uniform(size=(n, d)).astype("float32")
    y = np.random.choice(n_classes, size=n).astype(y_dtype)
    est = KerasClassifier(..., run_eagerly=run_eagerly)
    est.fit(X, y)
    est.score(X, y)
```

adriangb: I like that. Let's do that next PR.

adriangb:
I think this is ready to merge @stsievert ?

stsievert:
Yup, this looks good to me.

tests/test_input_outputs.py
adriangb: By the way @stsievert, if you want tests to run faster, you can do:
```python3
pip install pytest-xdist
pytest -n auto
```

PR #74, opened 2020-08-30
DOC: Mkdocs test

adriangb:


stsievert:
> I know we just implemented the RTD docs, but I also would like to float the idea of using mkdocs instead. From what I've seen a lot of state of the art libraries (in terms of Python deployment) like pydantic and poetry are using mkdocs. Apparently mkdocs now supports auto-generation of API docs: https://git>hub.com/mkdocs/mkdocs/wiki/MkDocs-Plugins#api-documentation-building. One thing I think mkdocs may be lacking is build-on-pr support (which is essential for a PR like this one)
>
> https://github.com/adriangb/scikeras/pull/73#issuecomment-683445483

I think the difference mostly comes down to Restructured Text (ReST) ReST or Markdown. I don't have a strong preference. Generally, I'm a fan of ReST for documentation because it can easily handle specialized formatting like warnings/notes, citations/references and argument lists.

Here's a relevant thread with some Python core devs chiming in: https://discuss.python.org/t/sphinx-rest-vs-markdown/2470/3

adriangb:
Thanks for the link to the interesting discussion!

My observation has been that mkdocs is able to generate nicer interfaces than what's possible on RTD, and having it hosted on github is also a big + (less systems). But I do understand that it has it's limitations.

https://myst-parser.readthedocs.io/en/latest/ this was linked in that discussion and looks interesting. It would be nice to be able to write some of the documentation in markdown (the simpler parts maybe) and some in rst (the more dynamic or complex parts).

adriangb:
I'm going to close this for now, I don't feel it's worth exploring any further, at least for now.

PR #85, opened 2020-09-26
MAINT: longer line length for black

stsievert:
It configures black to have the default line length of 88 characters instead of 79 characters.

I'm submitting this PR because I almost recommended this change multiple times in #66. I held back not to make that PR more complicated and make the diff even longer.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -158,16 +158,12 @@ def _linkcode_resolve(domain, info, package, url_fmt, revision):
     if not fn:
         return
 
-    fn = os.path.relpath(
-        fn, start=os.path.dirname(__import__(package).__file__)
-    )
+    fn = os.path.relpath(fn, start=os.path.dirname(__import__(package).__file__))
     try:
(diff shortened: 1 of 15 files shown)


adriangb:
Any reason not to stick with what we have? I feel like this is a personal preference thing. I'm not too opinionated about this stuff, and I think it's only 79 because that's what it was coming from TF. The fact that it is currently 79 means a pretty good reason is required to change it, even if it is only 79 for historical reasons.

stsievert:
> Any reason not to stick with what we have?

Yes. Reviewing PRs becomes more difficult with larger line lengths. Longer diffs are harder to review, almost by definition (more distance between relevant content). Increasing the black line length helps make shorter diffs, especially because black already emphasizes vertical white space. This got to me several times in #66 – I almost recommended this PR several times.

adriangb:
Hmm I mean I don't mind the change, and I guess you are doing most of the reviewing lately, so if this will make reviewing easier for you I'm happy to implement it.

It does look like linting failed. Mind fixing that?

Issue #79, opened 2020-09-22
Followups to #66


adriangb:
Possible followups from discussion in #66 

- [ ] Adding regex and/or globbing based parameter routing using a `__params_group` suffix.
- [x] Catching user errors when users compile their own models (https://github.com/adriangb/scikeras/pull/66#discussion_r492362306, done in #86)
- [ ] More user-friendly warnings when parameters are passed to an item that is not a class (https://github.com/adriangb/scikeras/pull/66#discussion_r492936838)
- [x] A bidirectional mapping between shorthand names for losses/metrics/optimizers and their functions/classes (https://github.com/adriangb/scikeras/pull/66#discussion_r492863842) Unidirectional mapping implemented in #88.
- [x] Consider making utilities public and refactor testing to remove tests for private functions.

stsievert:
Can you add making "Make some utilities public, and remove tests of private functions"?

adriangb:
I looked around, it looks like the only place we are testing "private" functions is within `test_utils.py`. It should be easy enough to replace them. Are there any specific utilities or private methods you think we should make public?

Regarding catching of user errors when compiling their models, would you be open to making this a warning instead of an error? I feel that it still achieves the goal of letting the user know they may have made a mistake while not going so far as to prohibiting users from doing things we don't approve of or haven't considered necessary.

stsievert:
> Are there any specific utilities or private methods you think we should make public?

I think the bidirectional map(s) between losses/optimizers/metrics should be public. Maybe `LabelDimensionTransformer` if users can (easily) customize BaseWrapper to use a different class. I'm not seeing why users would use any other utilities.

> making this a warning instead of an error? I feel that it still achieves the goal of letting the user know they may have made a mistake while not going so far as to prohibiting users from doing things we don't approve of or haven't considered necessary.

Generally, I tend to raise errors with helpful messages (or at least I try). Every time I write code to raise an exception I think "how could the user clear resolve this issue?" and include those actions in the error message (e.g,  https://github.com/adriangb/scikeras/pull/66#issuecomment-696750092).

adriangb:
> Maybe LabelDimensionTransformer if users can (easily) customize BaseWrapper to use a different class

[...]

PR #86, opened 2020-09-30
ENH: user friendly error if there is no loss function

adriangb:
As discussed in several previous issues. Checks off a task on #79.

This does not break any APIs because it does not check the `__init__` param. Instead it directly checks `tf.keras.Model.loss` after the user returns control over to SciKeras.

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -339,8 +339,9 @@ class BaseWrapper(BaseEstimator):
             raise ValueError(
                 "No valid loss function found."
                 " You must provide a loss function to train."
-                "\n\nPlease provide a loss function via the `loss` parameter or"
-                " compile your model with a loss function within your `model`"
+                "\n\nTo resolve this issue, do one of the following:"
+                "\n 1. Provide a loss function via the loss parameter."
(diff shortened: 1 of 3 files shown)


scikeras/wrappers.py
```diff
@@ -332,6 +332,20 @@ def _build_keras_model(self):
...
+            and not any(callable(l) or isinstance(l, str) for l in model.loss)
+        ):
+            raise ValueError(
+                "No valid loss function found."
+                " You must provide a loss function to train."
+                " Please provide a loss function via the `loss` parameter or"
+                " compile your model with a loss function within your `model`"
+                " model-building method."
+                " See https://www.tensorflow.org/api_docs/python/tf/keras/losses"
+                " for more information on Keras losses."
```
stsievert, line 346: Style nit: maybe add newlines so the user sees the following?

> No valid loss function found. You must provide a loss function to train.
>
> To resolve this issue, do one of the following:
> 1. Provide a loss function via the `loss` parameter.
> 2. Compile your model with a loss function inside the model-building method.
>
> See https://www.tensorflow.org/api_docs/python/tf/keras/losses Keras loss documentation.

PR #88, opened 2020-10-01
BUG/ENH: Data processing refactor

adriangb:
Closes #78, Closes #100, Closes #83.

This PR implements a sklearn transformer based interface for data processing that replaces `{pre,post}process_{X,y}`.

Issue #69, opened 2020-08-27
Implement `classes` param for partial fit


adriangb:
As per `MLPClassifiers` docs for `partial_fit` (and similar in other sklearn estimators):
> Classes across all calls to partial_fit. Can be obtained via np.unique(y_all), where y_all is the target vector of the entire dataset. This argument is required for the first call to partial_fit and can be omitted in the subsequent calls. Note that y doesn’t need to contain all labels in classes.

So the task for us is going to be:
- Accept classes argument
- Switch from LabelEncoder to OrdinalEncoder (which accepts a `classes` parameter).
- Patch that parameter into any other encoders (ex: `OneHotEncoder`).

I think Keras models support subsequent fit calls with new classes? But I'm not sure.

stsievert:
What changes would the user see from this change? The API will certainly change but what will that enable?

adriangb:
Currently, I'm not sure what happens when you do a partial fit where your first batch does _not_ contain all classes. There are two issues that we might run into:
1. How Keras/TF handle this. We need to do some testing here. This may be a dealbreaker.
2. How to handle our encoding/decoding in `preprocess_{X,y}`. Currently, everything is re-set for every call to `fit`, which means that predictions can only be made with the classes from the last call. Assuming we can figure out 1. above, we would need to figure out how to use the `classes` param to tell our encoders/decoders which classes to expect.

stsievert:
This issue is only relevant if a model changes the number of outputs based on the unique values in `y`, correct? I don't see where `classes` is needed with the MNIST dataset where the number of outputs is a known constant and hard-coded.

adriangb:
I think that typically you do something like: `Dense(n_classes_, activation="softmax")` to define your model output.

The issue here is that if `n_classes_` is calculated based on the data given in the first call to `partial_fit`, that data may not have all of the classes (depending on how the data was split).
For example:
```python3
y = [1, 2, 1, 2, 3]
y1 = y[:2]  # [1, 2]
y2 = y[2:]  # [1, 2, 3]

clf.partial_fit(X1, y1)  # uses Dense(2, activation="softmax")
clf.score(X1, y1)  # all good
clf.partial_fit(X2, y2)  # uh oh, now we have 3 classes! does Keras fail here? I'm not sure
clf.score(X2, y2)  # does this work even if the above failed? at the very least it's undefined behavior
```

stsievert:
[...]

PR #93, opened 2020-10-06
MAINT: simplify data cleaning logic

stsievert:
This PR cleans logic around initialization. Now, instead of `reset` parameter being passed to various functions, initialization is done once. Now, there's this branch:

``` python
def _initialize(self, X, y):
    # validate data, preprocess, etc
    return X, y, meta

if should_init:
    X, y, meta = self._initialize(X, y)
else:
    X, meta = self.preprocess_X(X)
    y = self.preprocess_y(y)
```

In addition, these simplifications were made:

* Fixes typo ("MultiOuputClassifier" → "MultiOutputClassifier")
* Simplification of `KerasClassifier.preprocess_y`. Now, convoluted if-statements are avoided.
* Uses `y.ndim` instead of `len(y.shape)`

TODO:

- [ ] Remove `scikeras.utils.type_of_target` pending https://github.com/adriangb/scikeras/pull/88/files#r499172803.


stsievert:
A couple of comments:

* I would have a flat namespace in `scikeras.utils` (so no `utils.transformers` namespace). I don't think the utils are complex enough to warrant more namespaces.
* I strongly recommend that you do not implement complex custom classes for your tests (i.e., `MultiOutputClassifier`). The tests are supposed to test the code in the `scikeras/` directory, not the code in the `tests/` directory.

scikeras/wrappers.py
```diff
@@ -467,35 +463,31 @@ def _check_array_dtype(arr):
...
+            n_features = X.shape[1]
+            meta.update(
+                {"X_dtype_": X.dtype, "X_shape_": X.shape, "n_features_in_": n_features}
+            )
+        if y is not None:
+            target_type = None if y is None else type_of_target(y)
+            meta.update(
+                {"y_dtype_": y.dtype, "y_ndim_": y.ndim, "target_type_": target_type}
+            )
+        return meta
```
adriangb, line 482: This seems to only get meta for `X` and `y`. Could it be renamed to `_get_data_meta` or something? Or are there plans to expand the functionality?

scikeras/wrappers.py
```diff
@@ -467,35 +463,31 @@ def _check_array_dtype(arr):
...
+        meta = {}
+        if X is not None:
+            n_features = X.shape[1]
+            meta.update(
+                {"X_dtype_": X.dtype, "X_shape_": X.shape, "n_features_in_": n_features}
+            )
+        if y is not None:
+            target_type = None if y is None else type_of_target(y)
+            meta.update(
+                {"y_dtype_": y.dtype, "y_ndim_": y.ndim, "target_type_": target_type}
```
adriangb, line 480: For `X` your are returning `"X_shape_"` but for `y` you are returning `"y_ndim_"`. Is this intentional? It makes sense to me, but I just want to double check.

scikeras/wrappers.py
```diff
@@ -591,6 +570,28 @@ def fit(self, X, y, sample_weight=None):
             X=X, y=y, sample_weight=sample_weight, warm_start=self.warm_start
         )
 
+    def _initialized(self):
+        return hasattr(self, "n_features_in_")
```
adriangb, line 574: Is `n_features_in_` the best attribute to check? We also have `is_fitted_`. I feel like we should always check the same attribute, and if that's not going to be `is_fitted_` we should get rid of `is_fitted_`.

scikeras/wrappers.py
```diff
@@ -888,84 +876,64 @@ def preprocess_y(
...
+            ),
+            "multiclass": make_pipeline(
+                Ensure2DTransformer(), OrdinalEncoder(dtype=np.float32),
+            ),
+            "multiclass-one-hot": FunctionTransformer(),
+        }
+        if is_categorical_crossentropy(loss):
+            encoders["multiclass"] = make_pipeline(
+                Ensure2DTransformer(), OneHotEncoder(sparse=False, dtype=np.float32),
+            )
```
adriangb, line 895: This is really clean, thank you.

scikeras/wrappers.py
```diff
@@ -888,84 +876,64 @@ def preprocess_y(
...
+                {"classes_": np.arange(0, y.shape[1]), "n_classes_": y.shape[1],}
+            )
+
+        if self.target_type_ not in ["binary", "multiclass", "multiclass-one-hot"]:
             raise ValueError(
                 f"Unknown label type: {self.target_type_}."  # To match errors used by sklearn
                 "\n\nTo implement support, subclass KerasClassifier and override"
                 " `preprocess_y` and `postprocess_y`."
                 "\n\nSee (TODO: link to docs) for more information."
             )
```
adriangb, line 929: This is a duplicate check/error from line ~897.

scikeras/wrappers.py
```diff
@@ -888,84 +876,64 @@ def preprocess_y(
...
+        encoder = self.target_encoder_
+
+        if self.target_type_ in ["binary", "multiclass"]:
+            meta.update(
+                {
+                    "classes_": encoder[1].categories_[0],
+                    "n_classes_": encoder[1].categories_[0].size,
+                }
+            )
         elif self.target_type_ == "multiclass-one-hot":
```
adriangb, line 918: We end up still keeping some `if/elif/else` logic here. We also have the same logic in `postprocess_y`. While I think this is cleaner, I like that in the old implementation there was a 1:1 map between the logic in `preprocess_y` and `postprocess_y`. Not a big deal, it's still clean, but I just wanted to point it out.

scikeras/wrappers.py
```diff
@@ -627,16 +628,22 @@ def _fit(self, X, y, sample_weight=None, warm_start=False):
...
+            X, meta = self.preprocess_X(X)
+            y, _ = self.preprocess_y(y)
+        self._check_output_model_compatibility(y)
+
+        if meta["n_features_in_"] != self.n_features_in_:
+            raise ValueError(
+                f"`X` has {meta['n_features_in_']} features, but this "
+                f"{self.__name__} is expecting {self.n_features_in_} "
+                f"features as input."
+            )
```
adriangb, line 646: This check is kind of lonely. Shouldn't we be checking all of `meta`'s keys (where it makes sense to do so)?

tests/multi_output_models.py
```diff
@@ -9,84 +9,85 @@
...
-                    meta.update(
-                        {
-                            "classes_": classes_,
-                            "n_classes_": [len(c) for c in classes_],
-                            "target_encoder_": target_encoder_,
-                            "model_n_outputs_": len(y),
-                            "n_outputs_": len(y),
-                        }
+    def preprocess_y(self, y):
+        reset = True
```
adriangb, line 19: Leftover I think, we can delete this and change the branch below.

tests/test_compile_kwargs.py
```diff
@@ -9,7 +9,7 @@
 from tensorflow.keras.models import Model
 
 from scikeras.wrappers import KerasClassifier
-from tests.multi_output_models import MultiOuputClassifier
```
adriangb, line 12: Thank you for the typo fix!

adriangb (commented):
Overall this looks great! Thank you for the PR.

adriangb:
> * I would have a flat namespace in `scikeras.utils` (so no `utils.transformers` namespace). I don't think the utils are complex enough to warrant more namespaces.

Sure, no problem changing that, we can do it after merging this PR into the other dev branch.

> * I strongly recommend that you do not implement complex custom classes for your tests (i.e., `MultiOutputClassifier`). The tests are supposed to test the code in the `scikeras/` directory, not the code in the `tests/` directory.

I totally agree, but one of the key features SciKeras promises to support that the old TF implementation did not is multi-input/output models. We need to ensure that our entire ecosystem (losses, paramter routing, class_weights, etc.) is compatible with that. How do we test all of that without implementing the tests the way I have? Previously it was straightforward because `MultiOutputClassifier` was in the `scikeras/` directory, but now that we are moving it out we should question the testing strategy. Do you have any ideas on how we can keep all of the tests for multi-input/output models that we have?

scikeras/wrappers.py
```diff
@@ -888,84 +876,64 @@ def preprocess_y(
...
-                y = encoder.transform(y)
+        encoders = {
+            "binary": make_pipeline(
+                Ensure2DTransformer(), OrdinalEncoder(dtype=np.float32),
+            ),
+            "multiclass": make_pipeline(
+                Ensure2DTransformer(), OrdinalEncoder(dtype=np.float32),
+            ),
+            "multiclass-one-hot": FunctionTransformer(),
+        }
```
adriangb, line 891: We instantiate new transformers and call `make_pipeline` every single time even if a couple lines down we ignore all of that work with `if not hasattr(self, "target_encoder_")`. I would move all of this within `if not hasattr(self, "target_encoder_").

scikeras/wrappers.py
stsievert: Yes. I checked; only `len(self.y_shape_)` is used. I debated doing the same to `X_shape_`, but feature shape is a lot more important than target shape.

adriangb:
@stsievert I have fixes for some of the minor nits above (https://github.com/adriangb/scikeras/pull/93#discussion_r500461198, https://github.com/adriangb/scikeras/pull/93#discussion_r500440493, https://github.com/adriangb/scikeras/pull/93#discussion_r500434979). Let me know if I can push it to this branch.

stsievert:
Yeah, feel free to push to this branch.

scikeras/wrappers.py
adriangb: I deleted `is_fitted_` and we can instead check `self._initialized()`.

scikeras/wrappers.py
adriangb: I agree. I think `y.ndim` is always going to be 1 or 2 as well. Thanks for clarifying.

scikeras/wrappers.py
adriangb: I moved this check to `preprocess_X`, where we were already checking `meta["X_dtype_"]`.

adriangb:
I think I ironed out some of the kinks.

Here are the things I'm not 100% on:
* We got rid of the `reset` argument in `preprocess_{X,y}` but we traded that for a new function `_get_meta`. Now it's hard to check meta params necessarily created in `preprocess_{X,y}` like `classes_` because we can't just say `hasattr(self, "classes_")` since that wouldn't differentiate between a `partial_fit` call (where you do want the check) or a second call to `fit` (where you don't want the check). Using a `reset` parameter is how sklearn handles this very same problem ([`BaseEstimator. _check_n_features`](https://github.com/scikit-learn/scikit-learn/blob/4ddd07967382c1aaea163b0ea8fd27682d3a5f67/sklearn/base.py#L352), [`BaseEstimator. _validate_data `](https://github.com/scikit-learn/scikit-learn/blob/4ddd07967382c1aaea163b0ea8fd27682d3a5f67/sklearn/base.py#L381)).
* The dictionary-based mapping of transformers is nice, but we still have `if/elif/else` logic that we can't get rid of, and the logic in `preprocess_y` no longer directly mirrors the logic in `postprocess_y`.

stsievert:
What's the purpose of `preprocess_y` now? Right now it chooses an encoder, collects metadata, and performs validation. Why not refactor `preprocess_y` into three functions?

It shouldn't be necessary to change the API to let the user specify the encoder. That might be a separate discussion.

> Using a reset parameter is how sklearn handles this very same problem

I don't think SciKeras needs that complication. I think the above will boil down to this code:

``` python
def _fit(self, X, y, warm_start=False):
    should_init = True
    if (self.warm_start or warm_start) and self._initialized():
        shoud_init = False
    if should_init:
        meta = self._get_meta(X=X, y=y)  # only get array metadata
        target_type = type_of_target(y)
        encoder = self._get_target_encoder(target_type).fit(y)
        vars(self).update({"target_type_": target_type, "target_encoder_": encoder, **meta})

        model, meta = self._build_keras_model()
        vars(self).update({"model_": model, **meta})

    # now, BaseWrapper has all the necessary attributes (X_dtype_, etc)
    self._check_targets(y)  # only raise errors
    self._check_features(X)  # only raise errors
    y = self.target_encoder_.transform(y)
    self._check_model_compat(X, y)  # only raise errors

    ...
```

I might be might be missing some functions/arguments/etc, but I think this dataflow should work.

adriangb:
> It shouldn't be necessary to change the API to let the user specify the encoder

I agree. But what about the `reset` parameter forces users to change the API? In #88 `MultiOutputClassifier` uses the reset parameter without any API changes ([link](https://github.com/adriangb/scikeras/blob/843192f58f54ec0ad5b94c4f696a3e337aff9b2a/tests/multi_output_models.py#L12)).

> What's the purpose of preprocess_y now? Right now it chooses an encoder, collects metadata, and performs validation

I'd say the 3 functions are:
* Collect metadata
* Perform validation.
* Prepare `{X,y}` for Keras.

The key points here are:
* Preparing `{X,y}` should not require implementing or using a Scikit-Learn style transformer. Users may want to just do `def preprocess_y(self, y, ...): return np.split(y, y.shape[1]), dict()`.
* Metadata collection and validation is up to the user. If they collect no metadata, they don't have to implement any validation. SciKeras should not assume any specific metadata outside of it's default implementations of `{pre,post}process_{X,y}`.


This means I would want to make `_get_meta` public and somehow get rid of `_get_target_encoder`.

And I still don't see how that API would allow setting and validating of attributes that are dependent on whatever processing is going on within `preprocess_{X,y}`, an implementation-specific example of this being `classes_`.

stsievert:
I've changed the example in https://github.com/adriangb/scikeras/pull/93#issuecomment-704557571 to reflect some of your concerns:

* `_get_meta` and `_get_target_transformer` are now public (and renamed to `fit_metadata` and `target_transformer` respectively).
* `fit_metadata` is passed an `encoder=` argument. This allows pulling `classes_`/etc from the encoder.

I'm not seeing the big barrier to using transformers. They require a little more structure, but it's pretty minimal:

``` python
class MultiOutputRegressor(KerasRegressor):
    def encode(self, y):  # aka preprocess
        self.foo = "bar"
        return [y[:, i] for i in range(y.shape[1])]

    def target_transformer(self, y):
        return FunctionTransformer(forward=encode, inverse_transform=np.column_stack)
```

I think this could easily replace `preprocess_{X,y}`, especially with some good documentation.

> And I still don't see how that API would allow setting and validating of attributes that are dependent on whatever processing is going on within preprocess_{X,y}

I would prefer to see the user override `target_transformer` if they have a stateful transform because it supports the `fit`/`transform` protocol. I think that `{pre,post}process_*` should be stateless or a static method (so in my example, `self.foo` isn't declared).

If you really want to support user doing *anything* in pre/postprocess, you have a good suggestion with renaming `reset`. In my example, I've added pre/postprocess back in and have added an attribute `BaseWrapper.initializing_`. That means this code is possible:

``` python
class NormalizedMultiOutputRegressor(KerasRegressor):
    def preprocess_y(self, y: ArrayLike) -> List[ArrayLike]:
        if self.initializing_:
            self.normalizers_ = [StandardScaler().fit(y[:, i] for i in range(y.shape[1])]
        return [norm.transform(y[:, i]) for i, norm in enumerate(self.normalizers_)]

    def postprocess_y(self, y: List[ArrayLike]) -> ArrayLike:
        y2 = [norm.inverse_transform(yi) for yi, norm in zip(y, self.normalizers_)]
        return np.column_stack(y2)
```

To be clear, I don't think `BaseWrapper.initializing_` should be used anywhere in the SciKeras codebase. I don't think it's necessary given that encoders are used in `KerasClassifier` and `KerasRegressor`. It's a nice documentation note (it will likely help extending SciKeras to multi input/output models), but that's all it should be. It's still not even necessary; here's some equivalent code:

<details>

``` python
class MultiNormalizer(Transformer):
    def fit(self, y: ArrayLike):
        self.normalizers_ = [StandardScaler().fit(y[:, i] for i in range(y.shape[1])]
        return self

    def transform(self, y: ArrayLike) -> List[ArrayLike]:
        return [norm.transform(y[:, i]) for i, norm in enumerate(self.normalizers_)]

    def inverse_transform(self, y: List[ArrayLike]) -> ArrayLike:
        y2 = [norm.inverse_transform(yi) for yi, norm in zip(y, self.normalizers_)]
        return np.column_stack(y2)

class NormalizedMultiOutputRegressor(KerasRegressor):
    def target_encoder(self, target_type):
        return MultiNormalizer()
```

</details>

adriangb:
That is great @stsievert, I _think_ it addresses a lot if not all of my concerns, but let me re-read with fresh eyes in the morning and give it a go at implementing into this branch and I'll loop back before pushing.

adriangb:
I generally really like the idea of encapsulating state within transformers, but I feel like if we're going to go that route, we should fully commit by completely replacing `{pre,post}process_{X,y}` with something like:
```python3
def _fit(self, X, y, warm_start=False):
    ...
    if should_init:
       # this branch could live inside self._initialize() or something, that's internal detail
        self.target_type_ = type_of_target(y)
        ...  # other collection of metadata like self.target_type_, so that all of it is available to get_target_encoder
        self.target_encoder_ = self.get_target_encoder().fit(y)
        self.input_encoder_ = self.get_input_encoder().fit(X)
        vars(self).update(**self.target_encoder.get_meta())  # do we want this? or are the encoders themselves now the metadata?
        ...
    y = self.target_encoder_.transform(y)
    X = self.input_encoder_.transform(X)
   ...
```
Where `get_target_encoder` is a public function with full access to all SciKeras metadata, so the default implementation would look something like:
```python3
class KerasClassifier:
    def get_target_encoder(self):
        # users can pass metadata to their TF and the TF decides which branch to use
        return KerasClassifierTransformer(loss=self.loss, type_of_target=self.target_type_)
        # or they could put that logic here and return different transformers
        tfs = {"multiclass": KerasClassifierMulticlassTF(loss=self.loss), "binary": KerasClassifierBinaryTF(loss=self.loss), ...)
        return tfs[self.target_type_]
```

And obviously we would completely delete `{pre,post}process_{X,y}`.

What do you think?

stsievert:
> we should fully commit by completely replacing {pre,post}process_{X,y}

:+1: yes, please. Your implementation looks good.

I only included `{pre,post}process_*` in https://github.com/adriangb/scikeras/pull/93#issuecomment-704682009 because https://github.com/adriangb/scikeras/pull/93#issuecomment-704579608 expressed a desire to keep that API.

> `vars(self).update(**self.target_encoder.get_meta())`  # do we want this? or are the encoders themselves now the metadata?

I think KerasClassifier should still have `classes_`. Here's the interface I'm thinking of for `fit_metadata`:

``` python
class BaseEstimator(BaseEstimator):
    def fit_metadata(self, X, y, target_encoder):
        meta = {"X_dtype_": X.dtype, ...}  # X metadata
        meta.update({"y_dtype_": y.dtype, ...})  # y metadata
        meta.update({"target_encoder_": target_encoder})
        return meta

class KerasClassifier(BaseWrapper):
    def fit_metadata(self, *args, **kwargs):
        meta = super().fit_metadata(*args, **kwargs)
        meta["classes_"] = meta["target_encoder_"][-1].classes_  # I think indexing is correct with Ensure2DTransformer
        return meta
```

> Where get_target_encoder is a public function w

A better name might be `target_encoder` and `feature_encoder` and with a property:

``` python
class BaseWrapper(BaseEstimator):
    @property
    def target_encoder(self):
        return make_pipeline(...)

    def _fit(...):
        ...
        if should_init:
            ...
            self.target_transformer_ = self.target_transformer.fit(y)
```

---

I've written this entire comment without a `feature_encoder`; I don't think that function is necessary or even required. I could only find a meaningful `preprocess_X` [in MultiOutputClassifier][2], and I think that can be taken care of outside of SciKeras with the "multiple output transformer" in https://github.com/adriangb/scikeras/pull/93#issuecomment-704682009 and this code:

``` python
from scikeras import KerasClassifier
from sklearn.pipeline import Pipeline
from sklearn.datasets import make_classification

classifier = KerasClassifier(...)
feature_transformer = MultiNormalizer()  # definition in "Multiple output transformer" of https://github.com/adriangb/scikeras/pull/93#issuecomment-704682009

estimator = make_pipeline(feature_transformer, classifier)
X, y = make_classification()
estimator.fit(X, y)
```

[2]:https://github.com/adriangb/scikeras/blob/c7609e89f29b2d9ddaa3923f23887b292973c0c3/tests/test_input_outputs.py#L52
[1]:https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html

adriangb:
> 👍 yes, please. Your implementation looks good.

Awesome will start working on it.

> [#93 (comment)](https://github.com/adriangb/scikeras/pull/93#issuecomment-704579608) expressed a desire to keep that API.

Yeah, sorry if I sound flip-floppy on that. If we _can_ do it all with TFs then that is a cleaner interface.


> I think KerasClassifier should still have `classes_`

I wouldn't want to hardcode retrieval of any specific parameters. I think we should either get them all or none. At the end of the day, the only direct consumer of those parameters is going to be `model_build_fn`.

> A better name might be `target_encoder` and `feature_encoder` and with a property:

Agreed.


> I think that can be taken care of outside of SciKeras

I actually don't think it can, because SciKeras can't accept a list of arrays or a dict of arrays as an input. If you do the transform before SciKeras, you'll get validation errors.

adriangb:
I have an initial working prototype. Still need to rework all of the tests that relied on the old API, so I'm not going to push just yet.

> I think KerasClassifier should still have classes_

Regarding this problem, here is what I ended up implementing:

```python3
class BaseScikerasDataTransformer(TransformerMixin, BaseEstimator, ABC):

    @abstractmethod
    def fit(self, X: np.ndarray) -> "BaseScikerasDataTransformer":
        pass
    
    @abstractmethod
    def transform(self, X: np.ndarray) -> Union[np.ndarray, List[np.ndarray], Dict[str, np.ndarray]]:
        pass

    @abstractmethod
    def inverse_transform(self, X: Union[List[np.ndarray], np.ndarray]) -> np.ndarray:
        pass
    
    @abstractmethod
    def get_meta_params(self) -> Dict[str, Any]:
        pass
```

```python3
class BaseWrapper(...):
    def _initialize(self, X, y):
        ...
        self.target_encoder_ = self.target_encoder.fit(y)
        target_meta = self.target_encoder_.get_meta_params()
        vars(self).update(**target_meta)
        self._meta.update(set(target_meta.keys()))
        self.feature_encoder_ = self.feature_encoder.fit(X)
        feature_meta = self.feature_encoder_.get_meta_params()
        vars(self).update(**feature_meta)
        self._meta.update(set(feature_meta.keys()))
       ...
```

This achieves the goal of letting users generate and set their own meta-params. In our default implementation, we use it to set all of the ones that used to be set in `preprocess_y`.

adriangb:
@stsievert  I just pushed here, heads up because I think you had some diverging changes, just in case you want to keep those.

stsievert:
>  I just pushed here

It looks like you've removed the commits with messages "Update tests.yaml" and "try to make test suite run." They're still on my machine, so I can force push those commits again. Let me know when I should do that.

adriangb:
No worries, I realized that after I commented, I edited my comment above but it seems like we crossed comments, sorry for the hassle 😄

adriangb:
I've got it down to just failures in `test_errors.py` corresponding to a bunch of `# TODO: validate here` in the code, I'll fix those later, but the gist of it is there if you want to take a look Scott.

The one thing I'll mention is that I re-added `reset` to `_validate_data`, but at least now it's a single function that uses that parameter, and it's private. I also removed the `meta` dict from `_validate_data` and instead set attributes directly. I don't care too much about the latter, so if you really think we should go with a `meta` dict we can do that, but to me it seemed like needless complexity. `_validate_data` now basically mirrors  `sklearn.base.BaseEstimator`.

scikeras/utils/_transformers.py
```diff
@@ -0,0 +1,262 @@
...
+
+class BaseKerasClassifierTargetTransformer(BaseScikerasDataTransformer):
+    """Base class for KerasClassifier target transformers.
+    """
+
+    @abstractmethod
+    def inverse_transform(
+        self, X: Union[List[np.ndarray], np.ndarray], return_proba: bool = False
+    ) -> np.ndarray:
+        pass
```
adriangb, line 77: All of these "Base" transformers are here to make the point that `KerasClassifier` expects a different signature for `inverse_transform`, but I think maybe it's overkill to have a base for each wrapper?

scikeras/utils/_transformers.py
```diff
@@ -0,0 +1,262 @@
...
+                self.y_dtype_, copy=False
+            )
+
+    def get_meta_params(self):
+        return {
+            "classes_": self.classes_,
+            "n_classes_": self.n_classes_,
+            "n_outputs_": self.n_outputs_,
+            "model_n_outputs_": self.model_n_outputs_,
+        }
```
adriangb, line 186: User gets to decide what attributes they want to return for setting in the main SciKeras wrapper

adriangb:
Tests are passing, coverage looks good and I think the interface is good too.

This should also be amenable to working with #69. I think we just propagate the `classes` param from `partial_fit` to `_fit` and then to the transformer's `__init__` as `categories` (rename it to match other sklearn TFs). And then since the default implementations are now using `OneHotEncoder` and `OrdinalEncoder` which take a `categories` parameter, we can just pass `categories=self.categories`.

I like it enough that I think we should merge and get back to #88 and discuss `scikeras.utils.type_of_target` there (which I think is the main issue left), but I'll leave this open until tomorrow in case you have any objections.

scikeras/wrappers.py
```diff
@@ -649,32 +634,15 @@ def _fit(self, X, y, sample_weight=None, warm_start=False):
             # because the predictions differ by a small margin).
             # To get around this, we manually delete these samples here
             zeros = sample_weight == 0
+            if zeros.sum() == zeros.size:
+                raise RuntimeError(
+                    "Cannot train because there are no samples"
+                    " left after deleting points with zero sample weight!"
```
stsievert, line 640: Shouldn't a ValueError be raised? That's what the documentation on [ValueError][2] and [RuntimeError][1] suggests. An incorrect value (the sample weight) was passed.

[1]:https://docs.python.org/3/library/exceptions.html#RuntimeError
[2]:https://docs.python.org/3/library/exceptions.html#ValueError

stsievert: Also, style nit: this might be a better message:

> No training samples had any weight; only zeros were passed in sample_weight={}. That means there's nothing to train on by definition, so training can not be completed.

scikeras/wrappers.py
```diff
@@ -1102,8 +928,8 @@ def predict_proba(self, X):
         # call the Keras model's predict
         outputs = self.model_.predict(X, **predict_args)
 
-        # join list of outputs into single output array
-        y = self.postprocess_y(outputs, return_proba=True)
+        # post process y
+        y = self.target_encoder_.inverse_transform(outputs, return_proba=True)
```
stsievert, line 932: The fact that `KerasClassifier.target_encoder_.inverse_transform` must accept a keyword arg `return_proba=True` will need documentation.

tests/multi_output_models.py
```diff
@@ -1,124 +1,101 @@
...
 from sklearn.preprocessing import FunctionTransformer, OneHotEncoder, OrdinalEncoder
 from tensorflow.python.keras.losses import is_categorical_crossentropy
 
-from scikeras.utils.transformers import Ensure2DTransformer
+from scikeras.utils import Ensure2DTransformer, KerasClassifierTargetTransformer
 from scikeras.wrappers import BaseWrapper, KerasClassifier
 
 
-class MultiOuputClassifier(KerasClassifier):
+class MultiOutputKerasClassifierTargetTransformer(KerasClassifierTargetTransformer):
```
stsievert, line 14: :-1: I think this class is too complex to go in the tests; it's about 70 lines, which is far more than a "simple class," which I consider to be 10 lines or less. I would consider this a simple class:

I do one of the following:

* Put this code in `scikeras/`, not `tests/`.
* Reduce the complexity of this class (by reducing the LOC required for this class).

You could also remove the encoders and have a simple multiple IO class:

``` python
class MultiOutputClassifier(KerasClassifier):
    @property
    def target_encoder(self):
        return FunctionTransformer(
            transform=lambda X: [X[:, 0], X[:, 1], X[:, 2:]],
            inverse_transform=np.concatenate,
        )
```

tests/test_input_outputs.py
```diff
@@ -24,6 +23,11 @@
 NUM_CLASSES = 2
 
 
+class MultiInputTransformer(FunctionTransformer):
+    def get_meta_params(self):
```
stsievert, line 27: Style nit: a better name might be `metadata`.

tests/test_utils_transformers.py
```diff
@@ -20,18 +20,26 @@ def test_basic_func(self, X):
...
-        X = np.array([1, 1])
         # Fit with 1D then pass 2D or 3D
-        # Raises ValueError
-        tf.fit(X)
-        with pytest.raises(ValueError, match="Expected a 1D array"):
+        X = np.random.uniform(size=100)
+        tf = Ensure2DTransformer().fit(X)
+
+        if X.ndim == 1:
+            with pytest.raises(ValueError, match="Expected a 1D array"):
```
stsievert, line 28: Should this error be raised? In the case this error is raised, there's only one meaningful dimension in `X_new`. Is this error raised when `X_new` has two dimensions and only one dimension is meaningful?

scikeras/wrappers.py
```diff
@@ -627,16 +613,15 @@ def _fit(self, X, y, sample_weight=None, warm_start=False):
...
 
         # Data checks
-        if warm_start and not hasattr(self, "n_features_in_"):
-            # Warm start requested but not fitted yet
-            reset = True
-        elif warm_start:
-            # Already fitted and warm start requested
-            reset = False
+        if not ((self.warm_start or warm_start) and self._initialized()):
+            X, y = self._initialize(X, y)
```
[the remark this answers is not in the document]
adriangb: I think this is my favorite:
``` python
        if not ((self.warm_start or warm_start) and self._initialized()):
            X, y = self._validate_data(X, y, reset=True)
            X, y = self._initialize(X, y)
        else:
            X, y = self._validate_data(X, y)
```
I'll go with that one

scikeras/wrappers.py
adriangb: Sure, I can change it. The thought process behind RuntimeError is that a value of all zeros for sample weight is not technically wrong, but if we delete all samples then that probably causes issues in Keras. I felt that it could be confusing to users if we say that their value for `sample_weight` is invalid if it is normally valid in `sklearn` (even if it is nonsensical).

scikeras/wrappers.py
adriangb: Agreed. My attempt to document this (other than docstrings, which I still need to work on) was creating all of those base transformers so that each wrapper has a defined signature for each one of the transformer functions.

scikeras/wrappers.py
adriangb: But thinking about it more, I think it's fine to raise a ValueError for nonsensical inputs, even if some sklearn estimators might succeed.

tests/test_input_outputs.py
adriangb: Changed name

tests/test_utils_transformers.py
adriangb: I think you originally introduced the error. I noticed it wasn't being tested (from coverage) and I thought I was fixing the test with this change. I could be wrong. I believe the idea is that if the user passes `X` as a column vector (specifically a column vector i.e. `(n_samples,)` and not `(1, n_samples, 1)`) to `fit` and then passes a >1D  `X` (ex: `(n_samples, 1)` or `(1, n_samples, 1)`) to `transform` this error should be raised. Does that make sense?

scikeras/utils/__init__.py
```diff
@@ -12,6 +12,19 @@
...
+    BaseKerasClassifierFeatureTransformer,
+    BaseKerasClassifierTargetTransformer,
+    BaseKerasRegressorFeatureTransformer,
+    BaseKerasRegressorTargetTransformer,
+    BaseScikerasDataTransformer,
+    Ensure2DTransformer,
+    KerasClassifierFeatureTransformer,
+    KerasClassifierTargetTransformer,
+    KerasRegressorFeatureTransformer,
+    KerasRegressorTargetTransformer,
```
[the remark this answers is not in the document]
adriangb: The goal here was to establish an "interface" for each transformer. Sadly, Python does not supports interfaces as intended here, so I settled on abstract classes.

`BaseScikerasDataTransformer` certainly achieves this because it forces implementation of the `get_metadata` method. 

`BaseKerasClassifierTargetTransformer` a bit less so because it cannot force signature compatibility (what we really want is to force the `return_proba` parameter).

`BaseKerasRegressorFeatureTransformer ` I agree is pretty useless, it's only there because it felt wrong to have a `BaseKerasClassifierTargetTransformer` without a corresponding interface/base class for KerasRegressor.

That said, do you still think we should keep only the 4 that you mentioned?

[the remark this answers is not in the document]
adriangb: Regarding the namespace, no problem, it'd be easy enough to delete the imports here and rename `_transformers.py` to `transformers.py`.

tests/multi_output_models.py
adriangb: I think this could be simplified like you indicate, but it would require removing tests that use this class + `{"sparse_categorical_crossentropy", "categorical_crossentropy"}`. I think we should be fine testing only with `"binary_crossentropy"`. We can save this more complex example for a notebook. What do you think?

adriangb: Turns out there was a single test that needed editing 😄 . The implementation is now only ~50 LOC, and most of it is boilerplate/docstrings.

adriangb:
Thanks for the review @stsievert. I addressed the easy points. Can we merge this and continue working on it in #88 since it seems that we both agree that generally this is the way forward?

adriangb (approved):

stsievert:
> Can we merge this and continue working on it in #88 since it seems that we both agree that generally this is the way forward?

I'd like to finish the review here first so existing conversations/threads can be resolved (and not lost). Is this PR blocking anything?

adriangb:
No worries then. I did want to start working on #89 and #69. My thought process was that the issues here are not very different than the issues in #88, but I guess yeah the conversations would be los. Let's finish up here first then.

tests/multi_output_models.py
adriangb: Simplified even further, looking at 27 lines of actual code. I think that should be acceptable.

tests/test_utils_transformers.py
adriangb: It looks like it was [here](https://github.com/adriangb/scikeras/blob/32a878cf8bc56cdf84d9679930cd7b329ad07d29/scikeras/utils/transformers.py#L17-L22).

I don't fully understand what the expected behavior is with `X.shape == (1, 400, 1)` since it that would result in `self.should_transform_ = False` from `fit` and hence this branch in `transform` would never be run. Am I missing something?

stsievert (commented):
Follow up to https://github.com/adriangb/scikeras/pull/93#discussion_r501864633 (can't reply because changes are outdated):

> The goal here was to establish an "interface" for each transformer. Sadly, Python does not supports interfaces as intended here, so I settled on abstract classes. 
>
> ...
>
> That said, do you still think we should keep only the 4 that you mentioned?

Yes. Python doesn't support interfaces, so those classes are meaningless. No user needs to subclass from them to make their class work with `BaseWrapper` – that's the main advantage of Python's duck-typing. I'd only implement classes that are used in SciKeras (`Ensure2DTransformer`, `ClassifierLabelEncoder`), probably with one class for inheritance (`BaseKerasTransformer` to say "SciKeras expects a `get_metadata` function").

tests/test_utils_transformers.py
adriangb: To me it makes sense as it is now (without any mention of `X.shape == (1, 400, 1)`). I'll leave it as it is now then.

scikeras/wrappers.py
adriangb: Ok I reduced the classes to just the ones discussed plus `KerasFunctionTransformer` which is just `FunctionTransformer` but implementing `get_metadata`. This new class replaces several of the old ones.

And yes, `return_proba` is specific to `ClassifierLabelEncoder` / `KerasClassifier`. I'm open to ideas if you think there's a better way to do it than having that argument.

adriangb:
> TODO:
> 
> * [ ]  Remove `scikeras.utils.type_of_target` pending https://github.com/adriangb/scikeras/pull/88/files#r499172803.

What do you think we should do with this issue?

scikeras/wrappers.py
stsievert: > Ok I reduced the classes to just the ones discussed

Thanks!

> KerasFunctionTransformer which is just FunctionTransformer but implementing get_metadata

I'd remove `KerasFunctionTransformer`. I think it would be more convenient to users to use arbitrary Scikit-learn transformers out-of-the-box rather than create a new class, instantiate it before using it.

I'm thinking of the code below alongside the documentation "metadata will be collected from `get_metadata` if the transformer implements that method."

``` python
class BaseWrapper(...):
    def _fit(self, X, y):
        ...
        meta = transformer.get_metadata() if hasattr(transformer, "get_metadata") else {}
```

adriangb: Sure, we can do a `hasattr` instead.

By the way, I went with the `get_meta` because that method already exists for `BaseWrapper`.

adriangb: Implemented the `metadata will be collected from get_metadata if the transformer implements that method.` in the docstrings as well as removing the class and using `hasattr` instead. Let's resolve this thread?

stsievert: Resolved

scikeras/wrappers.py
```diff
@@ -465,109 +464,89 @@ def _check_array_dtype(arr):
...
-            y : 2D numpy array with singular dimensions stripped
-                or 1D numpy array
+        Returns
+        -------
+        BaseKerasTransformer
+            Transformer implementing the BaseKerasTransformer
+            interface.
         """
-        return np.squeeze(np.column_stack(y))
+        raise NotImplementedError
```
stsievert, line 534: I think by default, BaseWrapper shouldn't do any data processing instead of raising a ValueError.

stsievert (commented):
Let's handle the discussion on `target_type` in #88. I think this PR is ready to merge after this comment.

scikeras/wrappers.py
adriangb: That's what I originally had, I'll revert, not a huge deal to me.

adriangb: Done

PR #105, opened 2020-10-11
MAINT: simplify logic in wrappers.py from #102

stsievert:
It's a follow up on #102. It incorporates those changes into `wrappers.py`. Specifically, simplifies the logic in that file and removes the unnecessary try/except blocks.

It also renames an error from RuntimeError to ValueError, the proper error because an array of the wrong shape is passed, and adds some more information to other error messages.


adriangb (approved):
This looks good to me, please go ahead and merge.

stsievert:
This is your repo; I'll leave merging to you. Also, feel free to revoke my commit access.

PR #123, opened 2020-11-05
ENH: Support autoencoders and other uses of BaseWrapper

### LINKED ISSUE #110: Target parameter required
In BaseWrapper.fit, the targets `y` are required.

https://github.com/adriangb/scikeras/blob/9f633f5c21a45b43986a8d779c3ef9e3005038bb/scikeras/wrappers.py#L556

 This poses challenges for models that don't require targets (like autoencoders).

adriangb:
Closes #110


scikeras/wrappers.py
```diff
@@ -674,22 +670,25 @@ def _fit(self, X, y, sample_weight, warm_start, epochs, initial_epoch):
...
-            X,
-            y,
-            sample_weight=sample_weight,
-            warm_start=warm_start,
-            epochs=epochs,
-            initial_epoch=initial_epoch,
-        )
+        # fit model if a target was given
+        if y is not None:
+            self._fit_keras_model(
```
stsievert, line 683: Why is model fitting performed only if a target is given?

adriangb: As far as I know, the only case in which you'd want to fit the Keras model without a target is when the `X` is a `tf.DataSet` (in which case the target is contained in `X`).

Can you think of any other situation where it would be valid to fit the actual model with `y=None`?

adriangb: And for `tf.DataSet`s, if we wanted to support them in the future, it would just be `if y is not None or isinstance(X, ...)`...

stsievert: Yes. Any unsupervised learning example. Examples include Scikit-learn's [TruncatedSVD][1], [NMF][2], and [PCA]. I'm sure Keras examples of those models can be created.

> As far as I know, the only case in which you'd want to fit the Keras model without a target is when the `X` is a `tf.DataSet`

So... `self._fit_keras_model` should be called if `y=None` (and `X` is a TF dataset)?

[1]:https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html#sklearn.decomposition.TruncatedSVD.fit
[2]:https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.NMF.html#sklearn.decomposition.NMF.fit
[3]:https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html#sklearn.decomposition.PCA.fit

adriangb: > So... `self._fit_keras_model` should be called if `y=None` (and `X` is a TF dataset)?

In that case, yes, that's how you call a Keras model with a `tf.DataSet`. But that's definitely for another PR.


> Yes. Any unsupervised learning example. Examples include Scikit-learn's [TruncatedSVD](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.TruncatedSVD.html#sklearn.decomposition.TruncatedSVD.fit), [NMF](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.NMF.html#sklearn.decomposition.NMF.fit), and [PCA]. 

Wouldn't those all be implemented similar to how I did here? That is, `BaseWrapper.fit` _is_ called with `fit(X=..., y=None)` but `tf.keras.Model` is never called with `y=None` since it is trained separately.

adriangb: I do think there is the possibility of cleaning up the need to have two models with shared layers if SciKeras could somehow handle multiple models and allow using one for the initial `fit` and another for `predict/transform`, but I don't know that there's a universal way to support that use case. If there is, maybe we can make a `KerasTransformer` class to make the implementation in this notebook/PR cleaner.

adriangb: After implementing this specific example, I imagine the following:

```python3
def get_fit_model() -> tf.keras.Model:
    """A regular SciKeras get_model"""
    ...

def get_tf_model(fit_model: tf.keras.Model) -> tf.keras.Model:
    """A regular SciKeras get_model"""
    # extract layers from fitted model?
    ...

def get_inverse_tf_model(fit_model: tf.keras.Model) -> tf.keras.Model:
    """A regular SciKeras get_model"""
    # extract layers from fitted model?
    ...


class KerasTransformer(BaseEstimator, TransformerMixin):

    def __init__(self, model, tf_model, inverse_tf_model):
       ...

    def fit(self, X, y):
        super().fit(X, y)
        self.tf_model_ = get_tf_model(self.model_)
        self.inverse_tf_model_ = get_tf_model(self.model_)
        return self
    
    def transform(self, X):
        # some preprocessing (same as predict, etc.)
        X_tf = self.tf_model_.predict(X)
        # some postprocessing? (new methods/transformers?)
        return X_tf
    
    def inverse_transform(self, X_tf):
        # some preprocessing? (new methods/transformers?)
        X = self.inverse_tf_model_.predict(X_tf)
        # some postprocessing (same as predict, etc.?)
        return X
```

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,517 @@
...
+   ],
+   "source": [
+    "autoencoder.fit(x_train, x_train)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "For our encoder layer, we need to build the model, but we do not necessarily want to fit it to any data (we are re-using the weights from `autoencoder`). SciKeras allows this by calling `fit(X=x_train)` or `fit(X=x_test)`. Note that there is no training happening (thus why it is okay to pass `x_test` to `fit`), rather SciKeras will just learn the input shapes and initialize the underlaying Keras model."
```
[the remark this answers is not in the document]
adriangb: How about:
```
Next we fit the encoder and decoder using training data and then round-trip through both using test data.
```
```python3
decoder.fit(encoder.fit_transform(x_train))
roundtrip_imgs = decoder.transform(encoder.transform(x_test))
```

scikeras/wrappers.py
adriangb: Yes:
```
ValueError: No gradients provided for any variable: ...
```

notebooks/Autoencoders.ipynb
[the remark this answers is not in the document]
stsievert: Yes, I think that works; "fit(X_test)" or "fit_transform(X_test)" is never called.

Why can't "roundtrip_imgs = autoencoder.fit(x_train).transform(x_test)" be used? I might be missing something (I haven't looked at the notebook recently).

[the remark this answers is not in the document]
adriangb: That would work for our test, but the whole point is to have the ability to encode/decode the images separately. That is, in testing `decoder.transform(encoder.transform(x_test)) == autoencoder.fit(x_train).transform(x_test)` but in "production" you'd be doing `encoded = encoder.transform(x_test)` and then do something with `encoded`, which you cannot do if all you have is `autoencoder`.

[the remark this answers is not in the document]
adriangb: I added:

```python3
encoded_imgs = encoder.transform(x_test)
space_saving = (1-encoded_imgs.astype(x_test.dtype).nbytes/x_test.nbytes)*100
```

Which exemplifies the above comment.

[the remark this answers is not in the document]
stsievert: > have the ability to encode/decode the images separately

Got it, thanks. That's the context I missed from the notebook.

scikeras/wrappers.py
stsievert: > ValueError: No gradients provided for any variable: ...

Right. Then maybe `y` should be required; the model always has an output (even it's an approximation of the input).

If not,  I think this code would satisfy both our concerns:

``` python
self._fit_keras_model(X=X, y=y or X, ...)
```

* `self._fit_keras_model` would always be called with `fit` (which I think is critical, and the reason for `fit`).
* It would work with autoencoders/unsupervised models where the input and output should match (an appropriate documentation note needs to added).
* It would allow setting `y=None`. If `X` is a dataset, it could be modified to the following:

``` python
self._fit_keras_model(X=X, y=(y or X) if not isinstance(X, tf.Dataset) else y, ...)
```

adriangb: What is the advantage to this proposed method over the current method in this PR? The `Dataset` part can be handled exactly the same way in both.

This also won't work with the current PR. We need to call `BaseWrapper.fit` and _not_ call `BaseWrapper.model_.fit` in order for initialization to happen before you can call `predict/transform`. I considered creating a separate method `initialize` that could be called instead of `BaseWrapper.fit`, but that would change the apparent API vs. sklearn transformers.

stsievert: > What is the advantage to this proposed method over the current method in this PR?

`self._fit_keras_model` is called when `y=None`. Why would you call `BaseWrapper.fit` and not fit a Keras model?

adriangb: To initialize internal stuff so that it is available for `predict`/`transform`, including:
1) Get/create Keras model. In this case it's a pretrained model used only for inference, not need to train it.
2) Instantiate feature transformers.
3) Learn input shapes and other data validation stuff.

stsievert: Which of the options below we trying to support?

* Allow for unsupervised models (e.g., `autoencoder.fit(X_train)`).
* Allow for pretrained models – `autoencoder.initialize(model=pretrained).predict(X_test)`.

If Keras requires that features `X` and targets `y` be passed to any model, I think forcing `y` to be specified is reasonable.

adriangb: Hmm. I think we should support both use cases.

For autoencoder based transformers specifically, unless we create a separate class like described in https://github.com/adriangb/scikeras/pull/123#discussion_r518244187, we need to have multiple models (like implemented in the tests/notebooks in this PR). For the non-training models (the models that are only used for inference), we need to be able to "initialize" everything by calling `fit` so that the sklearn interface works (`fit_transform`, etc.). Although it is an internal implementation, I do see how it could seem arbitrary to not call `model_.fit` if `y=None`.

Maybe we should:
1. Implement a `BaseWrapper.initialize` method that will pull together `BaseWrapper._initialize` and `BaseWrapper._validate_data` and make them public, with the purpose of supporting the general use case of pre-trained models. This would be a separate PR.
2. In these examples, implement an `AutoEncoderTransformer` class that looks something like https://github.com/adriangb/scikeras/pull/123#discussion_r518244187 but internally uses the `initialize` method. This would allow the external API to still fit into the sklearn API (`autoencoder.fit_transform(X)`) while avoiding the need for the `if y is not None` logic inside `BaseWrapper._fit`. I would not implement it as part of scikeras itself yet, since I am not convinced that a clean, general implementation is possible.

Does this sound good to you?

scikeras/wrappers.py
```diff
@@ -516,39 +516,37 @@ def _check_array_dtype(arr):
...
+                # sklearn.base.BaseEstimator._check_n_features
+                # since this method is not available in sklearn <= 0.22.0
+                if n_features_in_ != self.n_features_in_:
+                    raise ValueError(
+                        "X has {} features, but {} is expecting {} features "
+                        "as input.".format(
+                            n_features_in_, self.__class__.__name__, self.n_features_in_
+                        )
                     )
-                )
```
adriangb, line 548: This is all just an indentation difference

scikeras/wrappers.py
adriangb: @stsievert I implemented point (1) above. There are basically no changes to `_fit` now (aside from reshuffling logic into the public `initialize` method). This includes no `if y is not None` logic. While point (2) would be nice, I don't think it is required for this PR.

scikeras/wrappers.py
```diff
@@ -516,39 +516,37 @@ def _check_array_dtype(arr):
...
-
-        if y is None:
-            return X
         return X, y
```
adriangb, line 550: Always return both now, even if None (so return may be (None, None)).

scikeras/wrappers.py
```diff
@@ -611,13 +609,28 @@ def fit(self, X, y, sample_weight=None):
             initial_epoch=0,
         )
 
-    def _initialized(self):
+    def initialized(self):
```
adriangb, line 612: Make public

scikeras/wrappers.py
```diff
@@ -629,6 +642,31 @@ def _initialize(
...
+        y : array-like, shape `(n_samples,)` or `(n_samples, n_outputs)`, optional
+            True labels for `X`, by default None.
+        
+        Returns
+        -------
+        BaseModel
+            A reference to the BaseModel instance for chained calling.
+        """
+        self._initialize(X, y)
+        return self  # to allow chained calls like initialize(...).predict(...)
```
adriangb, line 668: `self._initialize(X, y)` returns modified `(X, y)`. We want to return `self` instead for the public method.

scikeras/wrappers.py
```diff
@@ -386,6 +372,22 @@ def _fit_keras_model(self, X, y, sample_weight, warm_start, epochs, initial_epoc
...
+            raise ValueError(
+                "No valid loss function found."
+                " You must provide a loss function to train."
+                "\n\nTo resolve this issue, do one of the following:"
+                "\n 1. Provide a loss function via the loss parameter."
+                "\n 2. Compile your model with a loss function inside the"
+                " model-building method."
+                "\n\nSee https://www.tensorflow.org/api_docs/python/tf/keras/losses"
+                " for more information on Keras losses."
+            )
```
adriangb, line 389: Moving this check to `_fit_keras_model` so that inference/pertained models do not require a loss function.

adriangb:
Copying the new integrated example from the notebook to here for visibility:
```python3
def get_fit_model(encoding_dim: int) -> keras.Model:
    """A regular SciKeras ``get_model``."""
    # This is our input image
    input_img = keras.Input(shape=(784,), name="input")
    # "encoded" is the encoded representation of the input
    encoded = keras.layers.Dense(encoding_dim, activation='relu', name="encoded")(input_img)
    # "decoded" is the lossy reconstruction of the input
    decoded = keras.layers.Dense(784, activation='sigmoid', name="output")(encoded)
    # This model maps an input to its reconstruction
    autoencoder_model = keras.Model(input_img, decoded)
    return autoencoder_model

def get_tf_model(fit_model: keras.Model) -> keras.Model:
    """Extract the input layer and encoded output from ``fit_model``."""
    return keras.Model(fit_model.get_layer("input").input, fit_model.get_layer("encoded").output)

def get_inverse_tf_model(fit_model: keras.Model, encoding_dim: int) -> keras.Model:
    """Create a new input layer and connect it to the fitted ``fit_model``."""
    encoded_input = keras.Input(shape=(encoding_dim,))
    output = fit_model.get_layer("output")(encoded_input)
    return keras.Model(encoded_input, output)


class KerasTransformer(BaseWrapper, TransformerMixin):
    """A class that enables transform and fit_transform.
    """

    def __init__(self, *args, tf_est: BaseWrapper = None, inv_tf_est: BaseWrapper = None, **kwargs) -> None:
        super().__init__(*args, **kwargs)
        self.tf_est = tf_est
        self.inv_tf_est = inv_tf_est


    def fit(self, X, sample_weight=None):
        # Fit the auto-encoder to predict X from X
        super().fit(X=X, y=X, sample_weight=sample_weight)
        # Copy encoder/decoder models to avoid modifying user input objects
        self.tf_est_ = copy.copy(self.tf_est)
        self.inv_tf_est_ = copy.copy(self.inv_tf_est)
        # Set params for encoder/decoder models
        self.tf_est_.set_params(fit_model=self.model_)
        self.inv_tf_est_.set_params(fit_model=self.model_, encoding_dim=self.encoding_dim)
        # Preprocess X
        # Fit tf_est_
        self.tf_est_.initialize(X=X)
        X_tf = self.tf_est_.predict(X=X)
        # Fit inverse TF
        self.inv_tf_est_.initialize(X_tf)
        return self

    def transform(self, X):
        X_tf = self.tf_est_.predict(X)
        return X_tf
    
    def inverse_transform(self, X_tf):
        X = self.inv_tf_est_.predict(X_tf)
        X = self.feature_encoder_.inverse_transform(X)
        return X


tf_est = BaseWrapper(model=get_tf_model, fit_model=None, verbose=0)
inv_tf_est = BaseWrapper(model=get_inverse_tf_model, fit_model=None, encoding_dim=None, verbose=0)
autoencoder = KerasTransformer(model=get_fit_model, tf_est=tf_est, inv_tf_est=inv_tf_est, loss="binary_crossentropy", encoding_dim=32, epochs=50)

autoencoder.fit(X=x_train)
roundtrip_imgs = autoencoder.inverse_transform(autoencoder.transform(x_test))
```

adriangb:
@stsievert can you take another look at this when you have a chance? Ty

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,632 @@
...
+    "id": "dBFFXT-__7KU"
+   },
+   "outputs": [],
+   "source": [
+    "from tensorflow import keras\n",
+    "\n",
+    "\n",
+    "def get_fit_model(encoding_dim: int) -> keras.Model:\n",
+    "    \"\"\"A regular SciKeras ``get_model``.\"\"\"\n",
+    "    # This is our input image\n",
```
stsievert, line 179: Style nit: I'd prefer more spacious and meaningful comments. I'd style this function like so:

``` python
def get_fit_model(encoding_dim: int) -> keras.Model:
    """Get an autoencoder.

    This autoencoder compresses a 28x28 image (784 pixels) down to a feature of length
    `encoding_dim`, and tries to reconstruct the input image from that vector.
    """
    input_img = keras.Input(shape=(784,), name="input")
    encoded = keras.layers.Dense(encoding_dim, activation='relu', name="encoded")(input_img)
    decoded = keras.layers.Dense(784, activation='sigmoid', name="output")(encoded)
    autoencoder_model = keras.Model(input_img, decoded)
    return autoencoder_model
```

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,632 @@
...
+   },
+   "source": [
+    "## Define Keras Model"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We will be defining a very simple autencoder. We define _three_ model building methods:\n",
```
stsievert, line 157: Is it necessary to define three functions? Could you do something like below?

``` python
def get_all_models(encoding_dim: int) -> Tuple[Model, Model, Model]
    input_img = keras.Input(shape=(784,), name="input")
    encoded = keras.layers.Dense(encoding_dim, activation='relu', name="encoded")(input_img)
    decoded = keras.layers.Dense(784, activation='sigmoid', name="output")(encoded)
    autoencoder_model = keras.Model(input_img, decoded)
    return autoencoder_model, encoded, decoded

def get_fit_model():
    return get_all_models()[0]

def get_model():
    return get_all_models()[1]

def get_inv_model():
    return get_all_models()[2]
```

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,632 @@
...
+    "        super().__init__(*args, **kwargs)\n",
+    "        self.tf_est = tf_est\n",
+    "        self.inv_tf_est = inv_tf_est\n",
+    "\n",
+    "\n",
+    "    def fit(self, X, sample_weight=None):\n",
+    "        # Fit the auto-encoder to predict X from X\n",
+    "        super().fit(X=X, y=X, sample_weight=sample_weight)\n",
+    "        # Copy encoder/decoder models to avoid modifying user input objects\n",
+    "        self.tf_est_ = copy.copy(self.tf_est)\n",
```
stsievert, line 231: deepcopy?

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,632 @@
...
+    "        self.inv_tf_est_ = copy.copy(self.inv_tf_est)\n",
+    "        # Set params for encoder/decoder models\n",
```
stsievert, line 233: Style nit: this comment adds no value; of course the function `set_params` sets parameters.

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,632 @@
...
+   "execution_count": 27,
+   "metadata": {
+    "colab": {
+     "base_uri": "https://localhost:8080/",
+     "height": 34
+    },
+    "id": "QQhtIWsMP4Ii",
+    "outputId": "1581384e-7012-4816-d4cf-a4190a30e50f"
+   },
+   "outputs": [
```
stsievert, line 303: Style nit: this is a lot of output, especially with the unnecessary printing of `Out[27]`. I think `autoencoder.fit(X=x_train);` or `_ = autoencoder.fit(X=x_train)` would remove the long output.

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,632 @@
...
+      "x_test.shape[1]: 784\nencoded_imgs.shape[1]: 32\nSpace saving: 95.92%\n"
+     ]
+    }
+   ],
+   "source": [
+    "encoded_imgs = autoencoder.transform(x_test)\n",
+    "print(f\"x_test.shape[1]: {x_test.shape[1]}\")\n",
+    "print(f\"encoded_imgs.shape[1]: {encoded_imgs.shape[1]}\")\n",
+    "space_saving = (1-encoded_imgs.astype(x_test.dtype).nbytes/x_test.nbytes)*100\n",
+    "print(f\"Space saving: {space_saving:0.2f}%\")"
```
stsievert, line 602: Style nit: I find it more valuable to say "the images can be compressed to 1/20th the size."

docs/source/tutorials.rst
```diff
@@ -15,4 +15,6 @@ The following are examples and notebooks on how to use SciKeras.
 
 * `Data Transformers <https://github.com/adriangb/scikeras/blob/master/notebooks/DataTransformers.ipynb>`_ - Use SciKeras data transformers to handle multi input, multi output and multidimensional Keras Model. `Run in Google Colab 💻 <https://colab.research.google.com/github/adriangb/scikeras/blob/master/notebooks/DataTransformers.ipynb>`_
 
+* `Autoencoder/Transformer Example <https://github.com/adriangb/scikeras/blob/master/notebooks/Autoencoders.ipynb>`_ - Learn how to make a Keras Autoencoder into a Scikit-Learn transformer. `Run in Google Colab 💻 <https://colab.research.google.com/github/adriangb/scikeras/blob/master/notebooks/Autoencoders.ipynb>`_
```
stsievert, line 18: Can these notebooks be rendered in HTML for the documentation? That way the user won't have to wait for GitHub to render them. (I think I've used [nb-sphinx][1] for this, but there are [other options][2]).

[2]:https://blog.jupyter.org/integrating-output-in-documentation-with-jupyter-sphinx-ecf569ddab85
[1]:https://nbsphinx.readthedocs.io/

scikeras/wrappers.py
```diff
@@ -516,39 +524,37 @@ def _check_array_dtype(arr):
...
-                )
-            # The following check is a backport from
-            # sklearn.base.BaseEstimator._check_n_features
-            # since this method is not available in sklearn <= 0.22.0
-            if n_features_in_ != self.n_features_in_:
-                raise ValueError(
-                    "X has {} features, but {} is expecting {} features "
-                    "as input.".format(
-                        n_features_in_, self.__class__.__name__, self.n_features_in_
+        if X is not None:
```
stsievert, line 527: Why would `X` be None?

tests/test_basewrapper.py
```diff
@@ -0,0 +1,73 @@
...
+        )
+        decoder = AutoEncoderTransformer(
+            model=decoder_model, **compile_params, **fit_params
+        )
+
+        # Training
+        autoencoder.fit(x_train, x_train)
+        roundtrip_imgs = decoder.fit_transform(encoder.fit_transform(x_test))
+        mse = mean_squared_error(roundtrip_imgs, x_test)
+        assert mse <= 0.05  # 0.05 is empirically determined
```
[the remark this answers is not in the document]
adriangb: > Nit: is this test ensuring that the data flow of autoencoder training works or that autoencoders have a good loss after training? 20 epochs seems like a lot of data for this relatively simple test.

It's currently doing both. Do you think we need to split it up into two or something?

> Nit: shouldn't this test make sure the input/output images are the same size?

Agreed, I'll add that.

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,632 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "id": "qCcyTjVkvoxR"
+   },
+   "outputs": [],
+   "source": [
+    "!python -m pip install git+https://github.com/adriangb/scikeras.git@master"
```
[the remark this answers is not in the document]
adriangb: Oh yes, good catch. This is just so that I can test stuff, good catch. I'll change it here, and in a seperate PR fix it in all of the other notebooks.

notebooks/Autoencoders.ipynb
adriangb: Unfortunately, they need to share the same layer/object. The simplest I can think of is:

```python3
encoding_dim = 32  # 32 floats -> compression of factor 24.5, assuming the input is 784 floats
input_img = keras.Input(shape=(784,))
encoded = keras.layers.Dense(encoding_dim, activation="relu")(input_img)
decoded = keras.layers.Dense(784, activation="sigmoid")(encoded)
autoencoder_model = keras.Model(input_img, decoded)
encoder_model = keras.Model(input_img, encoded)
encoded_input = keras.Input(shape=(encoding_dim,))
decoder_layer = autoencoder_model.layers[-1]
decoder_model = keras.Model(encoded_input, decoder_layer(encoded_input))

def get_fit_model():
    return autoencoder_model

def get_tf_model():
    return encoder_model

def get_inverse_tf_model():
    return decoder_model
```

But that looses the ability to set the encoding dimensions as a parameter.

notebooks/Autoencoders.ipynb
adriangb: Yep. It's `1/25`, not sure if you were just ball-parking very well or I'm doing some bad math here...

notebooks/Autoencoders.ipynb
adriangb: Went with `_ = autoencoder.fit(X=x_train)`

notebooks/Autoencoders.ipynb
adriangb: Removed

notebooks/Autoencoders.ipynb
adriangb: Good catch. `sklearn.base.clone` is what I wanted.

notebooks/Autoencoders.ipynb
adriangb: Much better, thank you.

scikeras/wrappers.py
adriangb: Must be a leftover from a previous iteration; I couldn't find a single situation where this would be required. I'll revert it. Good catch, thank you.

scikeras/wrappers.py
```diff
@@ -386,6 +372,29 @@ def _fit_keras_model(self, X, y, sample_weight, warm_start, epochs, initial_epoc
...
+            no_loss = True
+        if no_loss:
+            raise ValueError(
+                "No valid loss function found."
+                " You must provide a loss function to train."
+                "\n\nTo resolve this issue, do one of the following:"
+                "\n 1. Provide a loss function via the loss parameter."
+                "\n 2. Compile your model with a loss function inside the"
+                " model-building method."
+                "\n\nSee https://www.tensorflow.org/api_docs/python/tf/keras/losses"
```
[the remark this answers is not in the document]
adriangb: Can we just link to the [existing section in the docs](https://scikeras.readthedocs.io/en/latest/advanced.html#compilation-of-model)?

docs/source/tutorials.rst
adriangb: Do they take a long time for them to load? It takes ~1 sec for me. I'll consider it as part of a future documentation cleanup sprint (during which I also want to clean up all of the docstrings and document attributes).

scikeras/wrappers.py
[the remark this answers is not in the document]
adriangb: Added a link to that section of the docs.

notebooks/Autoencoders.ipynb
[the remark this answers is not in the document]
adriangb: Fixed here and in #129

adriangb:
Thank you for the review! Great catch not introducing https://github.com/adriangb/scikeras/pull/123#discussion_r524840544 to the codebase.

tests/test_basewrapper.py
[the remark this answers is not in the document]
stsievert: > It's currently doing both. Do you think we need to split it up into two, or just reduce the epochs and loosen the check?

20 epochs on the full MNIST dataset seems like overkill; it's not SciKeras's job to implement/test optimizers. If the model/loss changes after one `partial_fit` call, to me that says "the optimizer is hooked up correctly" and "the autoencoder data flow is supported."

notebooks/Autoencoders.ipynb
stsievert: Why not have a flag to determine the return value?

``` python
def get_coder(encoding_dim: int, out: str = "autoencoder") -> Tuple[Model, Model, Model]
    input_img = keras.Input(shape=(784,), name="input")
    encoded = keras.layers.Dense(encoding_dim, activation='relu', name="encoded")(input_img)
    decoded = keras.layers.Dense(784, activation='sigmoid', name="output")(encoded)
    autoencoder = keras.Model(input_img, decoded)

    returns = {"autoencoder": autoencoder, "encoder": encoded, "decoder": decoded}
    return returns[out]

# ...

fit_model = get_all_models(model=get_coder, model__out="autoencoder")
compress = get_all_models(model=get_coder, model__out="encoder")
expand = get_all_models(model=get_coder, model__out="decoder")
```

notebooks/Autoencoders.ipynb
stsievert: Yeah, I'm sure you're right. Resolving.

docs/source/tutorials.rst
stsievert: For me, it takes around 5 seconds (or more) for GitHub to render these notebooks (on two different browsers). That's a really long time in the web context. I'm sure the rendering time will grow with the notebook complexity (images/graphs, tables, etc).

I don't think it should be part of this PR either, just something I noticed.

tests/test_basewrapper.py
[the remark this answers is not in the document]
adriangb: Does 5 epochs sound reasonable? Or do you think we we should record the loss after 0 epochs (using `initialize()` or calling with a single datapoint) and then make sure it decreases after 1 epoch?

docs/source/tutorials.rst
adriangb: Noted. I opened #130 to track this.

notebooks/Autoencoders.ipynb
adriangb: Every time `get_coder` is called, it would return a different `encoded` object. There would be nothing tying the autoencoder model (which is what is being trained) to the `encoded` and `decoded` layers used for inference. Unless I am missing something?

adriangb: We could do something like:

```python3
def get_fit_model(encoding_dim: int, out: str) -> keras.Model:
    input_img = keras.Input(shape=(784,))
    encoded = layers.Dense(encoding_dim, activation="relu")(input_img)
    decoded = layers.Dense(784, activation="sigmoid")(encoded)
    autoencoder_model = keras.Model(input_img, decoded)
    encoder_model = keras.Model(input_img, encoded)
    encoded_input = keras.Input(shape=(encoding_dim,))
    decoder_layer = autoencoder_model.layers[-1]
    decoder_model = keras.Model(encoded_input, decoder_layer(encoded_input))

    returns = {"autoencoder": autoencoder_model, "encoder": encoder_model, "decoder": decoder_model}
    out_new = None
    while True:
        out_new = yield returns[out if out_new is None else out_new]

gen = get_fit_model(32, "autoencoder")

a = next(gen)
b = gen.send("decoder")
c = next(gen)
``` 

But that would require making SciKeras aware of generators, `send`, etc.

tests/test_basewrapper.py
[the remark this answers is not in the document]
adriangb: I implemented the above and it works. Since it is quite literally what you suggested ("if the model/loss changes after one partial_fit call...") I'm going to resolve this.

notebooks/Autoencoders.ipynb
stsievert: > Every time get_coder is called, it would return a different encoded and decoded object.

Only because Keras uses a different random state to create the encoder/decoder models, correct? Does Keras have scheme to initialize the models from the same random state?

adriangb: It's not really random state related. It's because the layers have to be connected, as in, connected in the tensorflow graph, so that when you train the auto-encoder the weights in the encoder/decoder are updated / the same. You could create a completely new model with the same architecture and just copy over the weights, but you would still need a reference to the autoencoder's trained layers (to get the weights) so it's no different really.

adriangb: In other words, if `==` was overloaded to compare architecture _and_ weights, what we need is:

`autoencoder_model[encode_layers] == encoder_model`

Which can most easily be achieved if we satisfy:

`autoencoder_model[encode_layers] is encoder_model`

adriangb: @stsievert let me know if I am still not being clear enough, I can make a small example or something

stsievert: Eh, I haven't commented because I don't think this is important enough to worry about.

adriangb: Got it, thank you. Do you plan on taking another look at the PR as a whole, or should I move it forward?

notebooks/Autoencoders.ipynb
stsievert: I'd still to prefer hide the comments in all of these functions; they don't add much information. In this comment, of course `keras.Input` is the input.

notebooks/Autoencoders.ipynb
```diff
@@ -0,0 +1,550 @@
...
+    "    \"\"\"\n",
+    "\n",
+    "    def __init__(self, *args, tf_est: BaseWrapper = None, inv_tf_est: BaseWrapper = None, **kwargs) -> None:\n",
+    "        super().__init__(*args, **kwargs)\n",
+    "        self.tf_est = tf_est\n",
+    "        self.inv_tf_est = inv_tf_est\n",
+    "\n",
+    "\n",
+    "    def fit(self, X, sample_weight=None):\n",
+    "        # Fit the auto-encoder to predict X from X\n",
```
stsievert, line 239: This function also has a fair amount of low-information comments.

notebooks/Autoencoders.ipynb
adriangb: Sorry for missing that!

notebooks/Autoencoders.ipynb
adriangb: Removed the comments

adriangb:
Got rid of those comments, which were nits anyway, so I think this is ready to move forward. Thank you for the reviews @stsievert !

Issue #70, opened 2020-08-27
BUG: can't grid search optimizers


adriangb:
It turns out that Keras optimizers are not picklable because they use lambdas inside functions 🤦 

This can be fixed by using `tf.keras.optimizers.serialize` and `tf.keras.optimizers.deserialize`.

I think that maybe we need to list out all of the keras object types `Model`, `optimizers`, `callbacks`, etc. and use `copyreg.pickle` to register how to pickle each one using their custom keras serialize/deserialize methods.

stsievert:
This issue might be worth reporting upstream, especially because pickle is a protocol. That means they could use `tf.keras.optimizers.{serialize, deserialize}` inside the pickle implementation (`__reduce__` or `__{set, get}state__`).

The post "[Pickle isn't slow, it's a protocol][1]" is somewhat related, and has a relevant example of using PyTorch serialization inside their Tensor class.

[1]:https://matthewrocklin.com/blog/work/2018/07/23/protocols-pickle

adriangb:
I totally agree with you that the answer is for TF to implement the pickle protocol using it's own serialize/deserialzie functions (which I presume are optimized). I already suggested something similar to the TF team at https://github.com/tensorflow/tensorflow/pull/39609. I could open PRs for every single place this could be implemented, but I'm not sure that that will make things go any faster.

That post smells just like this problem. It's unfortunate that the TF maintainers are a lot less nimble and flexible that PyTorch.

adriangb:
I left another comment in the TF PR. We'll see if we get a response. I'll give it a week or two and if not go ahead and implement the monkey patching here.

adriangb:
So the RFC was approved on the Keras end, but it did not cover the specific use case here, which includes `optimizers`, `losses`, `metrics` and `callbacks`. The pushback on these was mainly surrounding optimizers, given that there is no singular "official" API for serializing optimizers _with_ their weights. Instead, we would have to call `tf.keras.optimizers.Optimizer.get_weights/tf.keras.optimizers.Optimizer.set_weights`.

We may not be able to upstream this since there is no clear use case in Keras for serializing optimizers, callbacks, etc. _outside_ of a `Model`.

stsievert:
Thanks for commenting.

> given that there is no singular "official" API for serializing optimizers with their weights

[...]

PR #126, opened 2020-11-07
TST: ensure serialization doesn't affect training

stsievert:
It implements a test to ensure model serialization doesn't affect model training.

Currently, this test fails for `optimizer="adam"` and passes for `optimizer="sgd"`. I believe this is because Adam has some internal state; it has to track the second moment estimate/etc (which is also true for RMSProp, Adagrad, Adadelta, SGD with momentum, etc). SGD without momentum doesn't have any state – the only relevant state variables are the initialization parameters.

<details>
<summary>Partial fit test</summary>

``` python
def _weights_close(model1, model2):
    w1 = [x.numpy() for x in model1.model_.weights]
    w2 = [x.numpy() for x in model2.model_.weights]
    assert len(w1) == len(w2)
    for _1, _2 in zip(w1, w2):
        assert np.allclose(_1, _2, rtol=0.01)
    return True


def _reload(model, epoch=None):
    tmp = pickle.loads(pickle.dumps(model))
    assert tmp is not model
    if epoch:
        assert tmp.current_epoch == model.current_epoch == epoch
    return tmp


@pytest.mark.parametrize("optim", ["adam", "sgd"])
def test_partial_fit_pickle(optim):
    """
    This test is implemented to make sure model pickling does not affect
    training.

    (this is essentially what Dask-ML does for search)
    """
    X, y = make_regression(n_features=8, n_samples=100)

    m1 = KerasRegressor(model=dynamic_regressor, optimizer=optim, random_state=42)
    m2 = clone(m1)

    # Make sure start from same model
    m1.partial_fit(X, y)
    m2.partial_fit(X, y)
    assert _weights_close(m1, m2)

    # Train; make sure pickling doesn't affect it
    for k in range(4):
        m1.partial_fit(X, y)
        m2 = _reload(m2, epoch=k + 1).partial_fit(X, y)

        # Make sure the same model is produced
        assert _weights_close(m1, m2)

        # Make sure predictions are the same
        assert np.allclose(m1.predict(X), m2.predict(X))

```

</details>

**Reference issues/PRs**
This test comes from the implementation in https://github.com/adriangb/scikeras/issues/70#issuecomment-723151716.


adriangb:
This is awesome, thank you! I will review later tonight. The fix should be to bring in https://github.com/tensorflow/tensorflow/pull/39609 to SciKeras (at least until something like it is upstremed). I _think_ that will fix the issue. If not, we can implement custom serialization logic for optimizers.

adriangb:
It looks like SaveModel does not save optimizer weights. Yikes. We'll have to implement custom serialization for optimizers. I wonder if this is also the case for other stateful things, like callbacks (I think some keep internal track of epochs and such).

adriangb:
It looks like there is no easy (as in without digging deep, deep into TF) way to make this work for stateful optimizers: https://github.com/tensorflow/tensorflow/issues/44670.

stsievert:
> We'll have to implement custom serialization for optimizers.

Can you expand on https://github.com/tensorflow/tensorflow/pull/42749#issuecomment-723414141? I'm confused on the difference between "slots" and "weights".

It looks like `Optimizer._create_all_weights` ([source][2]) is used during initialization ([source][4]). Would that be useful?

[3]:https://github.com/tensorflow/tensorflow/blob/15ffb9d4fb9a440a5b355e3064f8737a435f872c/tensorflow/python/keras/optimizer_v2/adadelta.py#L96
[1]:https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/Optimizer
[2]:https://github.com/tensorflow/tensorflow/blob/15ffb9d4fb9a440a5b355e3064f8737a435f872c/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L761-L772
[4]:https://github.com/tensorflow/tensorflow/blob/15ffb9d4fb9a440a5b355e3064f8737a435f872c/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L596-L597

adriangb:
> Can you expand on [tensorflow/tensorflow#42749 (comment)](https://github.com/tensorflow/tensorflow/pull/42749#issuecomment-723414141)? I'm confused on the difference between "slots" and "weights".

It seems that the issue is related to optimizers that have "slots" (based on https://github.com/tensorflow/tensorflow/pull/42749#issuecomment-683964931). I think that "slots" are an implementation detail in TF; [this](https://www.tensorflow.org/api_docs/python/tf/keras/optimizers/Optimizer#slots_2) is the only information I could find in the docs. TL;DR I'm just as confused as you about this slots thing.

> It looks like `Optimizer._create_all_weights` ([source](https://github.com/tensorflow/tensorflow/blob/15ffb9d4fb9a440a5b355e3064f8737a435f872c/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L761-L772)) is used during initialization ([source](https://github.com/tensorflow/tensorflow/blob/15ffb9d4fb9a440a5b355e3064f8737a435f872c/tensorflow/python/keras/optimizer_v2/optimizer_v2.py#L596-L597)). Would that be useful?

Maybe. I'd have to spend some time with it, I'm pretty unfamiliar with the inner workings of this part of TF. I also am not sure that we should be depending on implementations of private functions within TF, even if we are able to figure out how it all works.

Let's at least wait until next week to see if there are any updates to https://github.com/tensorflow/tensorflow/issues/44670 or https://github.com/tensorflow/tensorflow/pull/42749. I also saw https://github.com/tensorflow/community/pull/281 where there is a question about this specific issue (https://github.com/tensorflow/community/pull/281#issuecomment-715596624).

adriangb:
A useful representation of the actual "state" that these optimizers have: https://github.com/geohot/tinygrad/blob/35a5c82a2aba057f89725a485aed994aba50eebe/tinygrad/optim.py

adriangb:
@stsievert should we close this since these tests were incorporated in #128 ?

stsievert:
I've added a note in the test to refer back to this PR for more detail (to provide more context). Merge this PR instead?

adriangb:
> I've added a note in the test to refer back to this PR for more detail (to provide more context). Merge this PR instead?

Awesome, thank you for adding the notes, they look great. I'll merge this as you suggest.

PR #127, opened 2020-11-07
ENH: allow explicit initialization of estimators without fitting for …

adriangb:
As discussed in https://github.com/adriangb/scikeras/pull/123#discussion_r518909777

Shuffles around existing logic into a new public `BaseWrapper.initialize` method that allows explicit initialization of SciKeras estimators without needing to call fit / doing any fitting of the actual Keras Model. This same method is called automatically when `fit` is called.


adriangb:
Making a draft since this may just end up being part of #123, depending on how much other work needs to happen in #123.

PR #128, opened 2020-11-09
MAINT/ENH: SaveModel based serialization

### LINKED ISSUE #70: BUG: can't grid search optimizers

### LINKED ISSUE #125: MAINT: move tf.keras.Model serialization to use SaveModel
Basically copying the implementation from https://github.com/tensorflow/tensorflow/pull/39609, which will also be nice because SciKeras can be a proving ground for the approach.

This will probably require a TF version bump to 2.3.0, since it needs the ability to use `Model.save("ram://...")`

adriangb:
Building upon the test added in #126.

This PR implements:
1. `SaveModel` based serialization.
2. General serialization for optimizers, metrics and losses.
3. Tests for serializing optimizers, metrics and losses.

Closes #126, closes #70, closes #125

### DIFF
--- a/scikeras/__init__.py
+++ b/scikeras/__init__.py
@@ -8,7 +8,7 @@ from tensorflow.python.keras import metrics as _metrics  # noqa
 
 
 _metrics.log_cosh = (
-    _metrics.metrics.logcosh
+    _metrics.logcosh
 )  # See https://github.com/tensorflow/tensorflow/pull/42097
 
 from scikeras import _saving_utils as saving_utils  # noqa
--- a/scikeras/__init__.py
+++ b/scikeras/__init__.py
@@ -4,14 +4,8 @@ __author__ = """Adrian Garcia Badaracco"""
 __version__ = "0.2.1"
 
 from tensorflow import keras
-from tensorflow.python.keras import metrics as _metrics  # noqa
 
-
-_metrics.log_cosh = (
-    _metrics.logcosh
-)  # See https://github.com/tensorflow/tensorflow/pull/42097
-
-from scikeras import _saving_utils as saving_utils  # noqa
+from scikeras import _saving_utils as saving_utils
 
 
 keras.Model.__reduce__ = saving_utils.pack_keras_model
--- a/scikeras/__init__.py
+++ b/scikeras/__init__.py
@@ -8,7 +8,7 @@ from tensorflow.python.keras import metrics as _metrics  # noqa
 
 
 _metrics.log_cosh = (
-    keras.metrics.logcosh
+    _metrics.metrics.logcosh
 )  # See https://github.com/tensorflow/tensorflow/pull/42097
 
 from scikeras import _saving_utils as saving_utils  # noqa
--- a/scikeras/__init__.py
+++ b/scikeras/__init__.py
@@ -5,10 +5,10 @@ __version__ = "0.2.1"
 
 from tensorflow import keras
 
-from scikeras import _saving_utils as saving_utils
+from scikeras import _saving_utils
 
 
-keras.Model.__reduce__ = saving_utils.pack_keras_model
-keras.losses.Loss.__reduce__ = saving_utils.pack_keras_loss
-keras.metrics.Metric.__reduce__ = saving_utils.pack_keras_metric
-keras.optimizers.Optimizer.__reduce__ = saving_utils.pack_keras_optimizer
+keras.Model.__reduce__ = _saving_utils.pack_keras_model
+keras.losses.Loss.__reduce__ = _saving_utils.pack_keras_loss
+keras.metrics.Metric.__reduce__ = _saving_utils.pack_keras_metric
+keras.optimizers.Optimizer.__reduce__ = _saving_utils.pack_keras_optimizer
--- a/scikeras/__init__.py
+++ b/scikeras/__init__.py
@@ -3,12 +3,18 @@
 __author__ = """Adrian Garcia Badaracco"""
 __version__ = "0.2.1"
 
-
-# Monkey patch log_cosh reference
-# See https://github.com/tensorflow/tensorflow/pull/42097
-# Will be removed whenever the
-# min supported version of tf incorporates the fix
-from tensorflow.python import keras  # noqa
+from tensorflow import keras
+from tensorflow.python.keras import metrics as _metrics  # noqa
 
 
-keras.metrics.log_cosh = keras.metrics.logcosh
+_metrics.log_cosh = (
+    keras.metrics.logcosh
+)  # See https://github.com/tensorflow/tensorflow/pull/42097
+
+from scikeras import _saving_utils as saving_utils  # noqa
+
+
+keras.Model.__reduce__ = saving_utils.pack_keras_model
+keras.losses.Loss.__reduce__ = saving_utils.pack_keras_loss
+keras.metrics.Metric.__reduce__ = saving_utils.pack_keras_metric
+keras.optimizers.Optimizer.__reduce__ = saving_utils.pack_keras_optimizer
--- a/scikeras/__init__.py
+++ b/scikeras/__init__.py
@@ -3,12 +3,12 @@
 __author__ = """Adrian Garcia Badaracco"""
 __version__ = "0.2.1"
 
-from tensorflow import keras
+from tensorflow import keras as _keras
 
 from scikeras import _saving_utils
 
 
-keras.Model.__reduce__ = _saving_utils.pack_keras_model
-keras.losses.Loss.__reduce__ = _saving_utils.pack_keras_loss
-keras.metrics.Metric.__reduce__ = _saving_utils.pack_keras_metric
-keras.optimizers.Optimizer.__reduce__ = _saving_utils.pack_keras_optimizer
+_keras.Model.__reduce__ = _saving_utils.pack_keras_model
+_keras.losses.Loss.__reduce__ = _saving_utils.pack_keras_loss
+_keras.metrics.Metric.__reduce__ = _saving_utils.pack_keras_metric
+_keras.optimizers.Optimizer.__reduce__ = _saving_utils.pack_keras_optimizer
--- a/scikeras/_saving_utils.py
+++ b/scikeras/_saving_utils.py
@@ -9,17 +9,16 @@ import numpy as np
 from tensorflow import io as tf_io
 from tensorflow import keras
 from tensorflow.keras.models import load_model
-from tensorflow.python.lib import io as tf_io
 from tensorflow.python.platform import gfile
 
 
-def walk(dir):
+def walk(top):
     """Recursive directory tree generator for directories.
 
-  Similar to os.walk, but supports TF RAM based filesystems.
+    Similar to os.walk, but supports TF RAM based filesystems.
 
-  Backported from TF Nightly > 2.3.1; this func is broken in <=2.3.1.
-  """
+    Backported from TF Nightly > 2.3.1; this func is broken in <=2.3.1.
+    """
 
     def _make_full_path(parent, item):
         # Since `os.path.join` discards paths before one that starts with the path
@@ -29,7 +28,6 @@ def walk(dir):
             return "".join([os.path.join(parent, ""), item])
         return os.path.join(parent, item)
 
-    top = str(Path(dir))
     listing = tf_io.gfile.listdir(top)
 
     files = []
@@ -61,13 +59,9 @@ def unpack_keras_model(packed_keras_model):
     b = BytesIO(packed_keras_model)
     with zipfile.ZipFile(b, "r", zipfile.ZIP_DEFLATED) as zf:
         for path in zf.namelist():
-            if not tf_io.file_io.file_exists_v2(
-                os.path.dirname(os.path.join(temp_ram_location, path))
-            ):
-                tf_io.file_io.recursive_create_dir_v2(
-                    os.path.dirname(os.path.join(temp_ram_location, path))
-                )
-            with gfile.GFile(os.path.join(temp_ram_location, path), "wb+") as f:
+            dest = os.path.join(temp_ram_location, path)
+            tf_io.gfile.makedirs(os.path.dirname(dest))
+            with gfile.GFile(dest, "wb") as f:
                 f.write(zf.read(path))
     return load_model(temp_ram_location)
 
--- a/scikeras/_saving_utils.py
+++ b/scikeras/_saving_utils.py
@@ -12,7 +12,7 @@ from tensorflow import keras
 from tensorflow.keras.models import load_model
 
 
-ram_prefix = "ram://"
+ram_prefix = f"ram:{os.path.sep}{os.path.sep}"
 
 
 def _temp_create_all_weights(self, var_list):
@@ -39,12 +39,12 @@ def _restore_optimizer_weights(optimizer, weights) -> None:
 def unpack_keras_model(packed_keras_model, optimizer_weights):
     """Reconstruct a model from the result of __reduce__
     """
-    save_folder = f"tmp/saving/{uuid()}"
+    save_folder = os.path.join("tmp", "loading", str(uuid()))
     temp_ram_location = ram_prefix + save_folder
     b = BytesIO(packed_keras_model)
     with zipfile.ZipFile(b, "r", zipfile.ZIP_DEFLATED) as zf:
         for path in zf.namelist():
-            dest = temp_ram_location + "/" + path
+            dest = os.path.join(temp_ram_location, path)
             tf_io.gfile.makedirs(os.path.dirname(dest))
             with tf_io.gfile.GFile(dest, "wb") as f:
                 f.write(zf.read(path))
@@ -57,7 +57,7 @@ def unpack_keras_model(packed_keras_model, optimizer_weights):
 def pack_keras_model(model):
     """Support for Pythons's Pickle protocol.
     """
-    save_folder = f"tmp/saving/{uuid()}"
+    save_folder = os.path.join("tmp", "saving", str(uuid()))
     temp_ram_location = ram_prefix + save_folder
     model.save(temp_ram_location)
     b = BytesIO()
--- a/scikeras/_saving_utils.py
+++ b/scikeras/_saving_utils.py
@@ -33,11 +33,11 @@ def unpack_keras_model(packed_keras_model, optimizer_weights):
     """Reconstruct a model from the result of __reduce__
     """
     save_folder = f"tmp/saving/{id(packed_keras_model)}"
-    temp_ram_location = os.path.join(ram_prefix, save_folder)
+    temp_ram_location = ram_prefix + "/" + save_folder
     b = BytesIO(packed_keras_model)
     with zipfile.ZipFile(b, "r", zipfile.ZIP_DEFLATED) as zf:
         for path in zf.namelist():
-            dest = os.path.join(temp_ram_location, path)
+            dest = temp_ram_location + "/" + path
             tf_io.gfile.makedirs(os.path.dirname(dest))
             with tf_io.gfile.GFile(dest, "wb") as f:
                 f.write(zf.read(path))
@@ -54,7 +54,7 @@ def pack_keras_model(model):
     """Support for Pythons's Pickle protocol.
     """
     save_folder = f"tmp/saving/{id(model)}"
-    temp_ram_location = os.path.join(ram_prefix, save_folder)
+    temp_ram_location = ram_prefix + "/" + save_folder
     model.save(temp_ram_location)
     b = BytesIO()
     with zipfile.ZipFile(b, "w", zipfile.ZIP_DEFLATED) as zf:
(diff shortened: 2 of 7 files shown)


adriangb:
Look like this is only working on `tensorflow>=2.4.0rc1` at the moment. The RAM filesystem is not fully functional in `2.3.1`, which is the latest stable release. This will probably have to wait until `2.4.0` then.

adriangb:
So tests are mostly working, but I'm seeing `SavedModels saved from Tensorflow V1 or Estimator (any version) cannot be loaded with node filters` now, and I'm not able to reproduce locally...

tests/test_serialization.py
```diff
@@ -123,3 +124,117 @@ def test_run_eagerly():
...
+    assert tmp is not model
+    if epoch:
+        assert tmp.current_epoch == model.current_epoch == epoch
+    return tmp
+
+
+@pytest.mark.parametrize(
+    "optim", ["adam", "sgd", keras.optimizers.Adam(), keras.optimizers.SGD(),],
+)
+def test_partial_fit_pickle(optim):
```
stsievert, line 149: Does this test pass now for all optimizers?

scikeras/__init__.py
```diff
@@ -3,12 +3,18 @@
...
+)  # See https://github.com/tensorflow/tensorflow/pull/42097
 
-keras.metrics.log_cosh = keras.metrics.logcosh
+from scikeras import _saving_utils as saving_utils  # noqa
+
+
+keras.Model.__reduce__ = saving_utils.pack_keras_model
+keras.losses.Loss.__reduce__ = saving_utils.pack_keras_loss
+keras.metrics.Metric.__reduce__ = saving_utils.pack_keras_metric
+keras.optimizers.Optimizer.__reduce__ = saving_utils.pack_keras_optimizer
```
stsievert, line 20: Shouldn't these lines go in the definitions of Model/Loss/etc?

tests/test_serialization.py
adriangb: Yep!

scikeras/__init__.py
adriangb: Do you mean in the class def within Tensorflow? Maybe I'm not understanding...

stsievert: Oh, I see. These are *Keras* classes, not SciKeras classes. Then why not put this into a Keras RFC?

adriangb: Right now, these things are very hacky. The optimizer hack is never going to make it past any sort of review (I'm not even sure it works for 100% of cases; it relies on private methods). `Model.__reduce__` does a bunch of zip file stuff that is also pretty hacky, and really should be replaced with TF utilities that need to be implemented on the C++ backend (I think) as discussed in the TF PR (https://github.com/tensorflow/tensorflow/pull/39609#issuecomment-690544525).

adriangb:
> So tests are mostly working, but I'm seeing `SavedModels saved from Tensorflow V1 or Estimator (any version) cannot be loaded with node filters` in CI runs now, and I'm not able to reproduce locally...

Looks like the problem was something related to collisions of the temporary ram folders. Fixed by using a uuid instead of the id(object).

tests/test_serialization.py
adriangb: Working now except on Windows. I'll need to get a Windows computer to test with.

scikeras/_saving_utils.py
```diff
@@ -0,0 +1,155 @@
...
+
+    See https://tensorflow.org/api_docs/python/tf/keras/optimizers/Optimizer#slots_2
+    """
+    self._create_all_weights_orig(var_list)
+    try:
+        self.set_weights(self._restored_weights)
+    except ValueError:
+        # Weights don't match, eg. when optimizer was pickled before any training
+        pass
+    del self._restored_weights
```
stsievert, line 66: Suggested change:

``` diff
-    del self._restored_weights
+    delattr(self, "_restored_weights")
```

scikeras/_saving_utils.py
```diff
@@ -0,0 +1,155 @@
...
+    self._create_all_weights = self._create_all_weights_orig
+
+
+def _restore_optimizer_weights(optimizer, weights) -> None:
+    optimizer._restored_weights = weights
+    optimizer._create_all_weights_orig = optimizer._create_all_weights
+    optimizer._create_all_weights = MethodType(_temp_create_all_weights, optimizer)
```
stsievert, line 73: Why is `MethodType` required? Does it have the same affect as [functools.partial][1]?

[1]:https://docs.python.org/3.8/library/functools.html#functools.partial

scikeras/__init__.py
stsievert: I think there's two levels for RFCs: the goal and how to achieve that goal.

This PR shows the goal is feasible and practical, and serves as a prototype (`_restore_optimizer_weights` is clearly a rough prototype). It's clear what needs to be implemented in a cleaner way. Inside TF, there's no downside to depending on private functions.

scikeras/_saving_utils.py
adriangb: MethodType binds the method to the live object: https://stackoverflow.com/a/37455782/6582418

scikeras/__init__.py
adriangb: I agree that this PR shows that it's feasible, but a (similar) thing is already at https://github.com/tensorflow/tensorflow/pull/39609.

I guess, let me just ask, what steps do you suggest be taken? Open a new RFC or PR the optimizer weights part specifically?

scikeras/_saving_utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+def pack_keras_model(model):
+    """Support for Pythons's Pickle protocol.
+    """
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        model.save(tmpdirname)
+        b = BytesIO()
+        with tarfile.open(fileobj=b, mode="w:gz") as tar:
+            tar.add(tmpdirname, arcname=os.path.sep)
+        b.seek(0)
```
adriangb, line 55: @stsievert how do you feel about saving to a temp directory? It works (that's the best thing I can say about it). But it seems wrong to me to serialize to disk and then load from disk to memory.

stsievert: Yeah, I wouldn't write to disk. Pythons io module has in-memory file pointers with StringIO and BytesIO. Would those work?

scikeras/__init__.py
stsievert: I might open a new PR to TF after this PR is merged. Or maybe add it to that PR with an explaining comment if the implementation is simple.

If they want an RFC they'll call for it.

scikeras/_saving_utils.py
adriangb: That is what we're using here. The pickle transfer happens as bytes. The issue is that _Keras/TF_ can't save to `BytesIO` or `StringIO`. That is:
```python
model.save("some/directory/on/disk")   # works
model.save(BytesIO())  # does not work
```
So what we are doing here is:
1. Save TF model to a temporary folder on disk.
2. Load from that temp dir into a `BytesIO` object.
3. Wrap the `BytesIO` object in Numpy and pickle that.

scikeras/__init__.py
adriangb: I have no problem opening a hacky PR in TF just for visibility. Just to be clear, we're talking specifically about the optimizer weights part right (not about Metrics or Model)? And specifically, about the hack to restore the weights for `Adam` and other optimizers that use slots, _not_ about implementing `__reduce__` for all optimizers?

scikeras/_saving_utils.py
stsievert: Why wouldn't pickle work to serialize a Keras model? I thought that's the point of https://github.com/tensorflow/tensorflow/pull/39609.

adriangb: This _is_ the same implementation as [tensorflow/tensorflow#39609](https://github.com/tensorflow/tensorflow/blob/d6ef36c9b8e91de97e883abda0b54181c798934b/tensorflow/python/keras/engine/training.py#L2276). The only difference is that there I am using TF's `ram://` filesystem instead of writing to disk (i.e. write to disk-like thing in RAM, then load that into BytesIO). But that doesn't work on Windows, hence why I am considering writing to actual disk here. We could stop supporting Windows in which case can implement things exactly like tensorflow/tensorflow#39609.

That said, obviously the goal is for this to eventually end up in TensorFlow itself via tensorflow/tensorflow#39609 or tensorflow/tensorflow#39609 + other PRs at which point we'd simply delete the implementation from SciKeras.

stsievert: > But [ram://] doesn't work on Windows,

That sounds like an upstream issue. I would let it be resolved upstream. The con is writing to disk (which is slow, especially for models like ResNet52 that are 100s of MB), and the pro (Windows support) is not that large. I would use the `ram://` filesystem.

adriangb: And drop Windows support for all of SciKeras?

scikeras/_saving_utils.py
stsievert: Got it. It doesn't serialize (/pickle/curry) the optimizer, it passes the memory address.

I think a comment would help clarify (which will almost certainly save time in the future).

scikeras/_saving_utils.py
stsievert: Let's say that only the RAM filesystem is supported. When will Windows users run into difficulties Linux/macOS users can overcome?

How many LOC are required to support both RAM filesystems and disk filesystems in this PR?

How soon do you think the RAM filesystem will support Windows?

adriangb: Windows users will run into pickle errors when using GridSearch, ensembles, etc.

I'm testing right now, I think it's actually only going to be like an extra 15 LOC to support both.

I'd guess TF 2.6.0 or something, so like 6 months?

stsievert: > I'm testing right now, I think it's actually only going to be like an extra 15 LOC to support both.

That's not too bad.

adriangb: It's up, and tests are passing.

scikeras/_saving_utils.py
```diff
@@ -0,0 +1,137 @@
...
+from tensorflow.keras.models import load_model
+
+
+def _get_temp_folder():
+    if os.name == "nt":
+        # the RAM-based filesystem is not fully supported on
+        # Windows yet, we save to a temp folder on disk instead
+        return tempfile.mkdtemp()
+    else:
+        return f"ram://{tempfile.mkdtemp()}"
```
adriangb, line 22: @stsievert this is the "secret sauce" to support the ram based filsystem on *nix and disk based on windows

scikeras/_saving_utils.py
stsievert: A point of clarification: I think serialization really only presents an issue when serialization occurs in the middle of training (because before this PR, optimizer weights were not preserved).

Scikit-learn does not serialize the model *during training* in their grid/random search, though they do serialize before/after with the Joblib backend. Dask-ML/Ray do serialize during training because they have to call `partial_fit` on different workers.

I'm not too familiar with ensemble methods. I imagine they do not serialize *during training,* especially if each model lives on it's own process.

adriangb: I think that's a good clarification.

The optimizer hack is technically separate from the move from the old `Model` serialization code (an assortment of Keras methods that is not officially supported by TF/Keras) to this code (which uses `SaveModel` as the backend).

The move to SaveModel was going to happen at some point because that's what Keras is going to have more support for going forward and that's what will eventually be upstreamed.

That said, from an implementation perspective, these two things are pretty intermingled, and since they're private implementation details anyway, it makes sense to make both changes in the same PR.

scikeras/_saving_utils.py
adriangb: Added a comment.

scikeras/_saving_utils.py
stsievert: > That said, from an implementation perspective, these two things are pretty intermingled, and since they're private implementation details anyway, it makes sense to make both changes in the same PR.

How many LOC do you estimate will need to be added to implement `optimizer.__reduce__` in https://github.com/tensorflow/tensorflow/pull/39609? I'm not talking about the tests, only the implementation.

If it's a pretty simple implementation, I think the `optimizer.__reduce__` implementation should go in that PR with comment explaining the change. I'm not sure if the TF team will accept the addition – in the minutes (https://github.com/tensorflow/community/pull/286#issuecomment-714793415), Francois said:

> * makes sense for models, not so much for optimizers (since state needs to be maintained, we only serialize configs)
> * ...  
> * we should not implement at the moment for objects (callbacks, optimizers, metrics, losses with lambdas, etc.) where we only promise to save the config, not the internal state (as this breaks guarantees and API safety; state is usually retrieved via different API calls; results of these can be used to set state too)

adriangb: I was referring to putting them together in _this_ PR. But I agree, currently https://github.com/tensorflow/tensorflow/pull/39609 is in a state of POC, so adding code that maybe won't make it into the final version in order to chart/scope the project in general is probably a good idea.

stsievert: I'd be inclined to reach out to the TF team to hear what they have to say; they had some not-mild pushback on serializing optimizer state. I'd be inclined to reach out to the TF team after this PR is firmed up.

adriangb: I _think_ the pushback wasn't because "it shouldn't be done" but rather because "it's not implemented cleanly in TF, and we don't want a complex `__reduce__` implementation". But yeah I'll follow up after this PR.

scikeras/__init__.py
stsievert: > we're talking specifically about the optimizer weights part right

My concern is Keras having a working serialization method. I think Models and optimizers are most important use cases (plus all metrics I can think of are stateless past some rolling average).

I'm not sure what the difference is between "restoring weights for Adam" and "implementing `__reduce__`." Do they both accomplish a working serialization method? Do they have a different implementation?

adriangb: Yes, they are fundamentally two distinct things.

In this PR, we are fixing the weight restoration bug by implementing some hacks within `Model.__reduce__`. I reported the bug and showed how this hack works in https://github.com/tensorflow/tensorflow/issues/44670 in hopes that that will make it easier for them to fix the bug. To actually fix the bug, one would have to implement an actual fix within the TF `SaveModel` ecosystem. I tried and failed to do that. Thus I don't see that there's any further PRs I can submit to TensorFlow.

As for implementing `Model.__reduce__`, I think we should update the existing PR with whatever implementation we land on here, sans the hack to restore optimizer weights since that is functionally unrelated.

I hope this makes sense! I know it's a convoluted topic.

scikeras/__init__.py
```diff
@@ -3,12 +3,12 @@
...
 __version__ = "0.2.1"
 
+from tensorflow import keras
 
-# Monkey patch log_cosh reference
-# See https://github.com/tensorflow/tensorflow/pull/42097
-# Will be removed whenever the
-# min supported version of tf incorporates the fix
-from tensorflow.python import keras  # noqa
+from scikeras import _saving_utils as saving_utils
```
stsievert, line 8: Why are you making `saving_utils` a public module? Or, why should the end-user use any function in `saving_utils`?

scikeras/_saving_utils.py
```diff
@@ -0,0 +1,140 @@
...
+                    zf.writestr(os.path.relpath(dest, temp_dir), f.read())
+                tf_io.gfile.remove(dest)
+    b.seek(0)
+    return (
+        unpack_keras_model,
+        (np.asarray(memoryview(b.read())), model.optimizer.get_weights()),
+    )
+
+
+def unpack_keras_optimizer(opt_serialized, weights):
```
stsievert, line 99: Why does can't `_create_all_weights` be a regular function? It'd be a little cleaner if this all moved to the `Optimizer` class.

<details>
<summary>Draft implementation</summary>
I think this is a little cleaner; some functions look like they should be class methods. This implementation might be missing some (possibly important) details.

``` python
def _unpack_optimizer(opt_serialized, weights):
    opt = keras.optimizers.deserialize(opt_serialized)
    opt._restore_weights(weights)
    return opt

class Optimizer(...):
    def __reduce__(self):
        weights = self.get_weights()
        opt_serialized = keras.optimizers.serialize(optimizer)
        return unpack_keras_optimizer, (opt_serialized, weights)

    def _restore_weights(self, weights):
        self._restored_weights = weights
        self._create_all_weights_orig = self._create_all_weights
        self._create_all_weights = MethodType(_temp_create_all_weights, self)
```

</details>

scikeras/_saving_utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+def pack_keras_optimizer(optimizer: keras.optimizers.Optimizer):
+    """Support for Pythons's Pickle protocol in Keras Optimizers.
+    """
+    opt_serialized = keras.optimizers.serialize(optimizer)
+    weights = optimizer.get_weights()
+    return unpack_keras_optimizer, (opt_serialized, weights)
+
+
+def unpack_keras_metric(metric_serialized):
```
stsievert, line 115: These functions for metric/loss don't do much.

scikeras/_saving_utils.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def unpack_keras_model(packed_keras_model, optimizer_weights):
+    """Reconstruct a model from the result of __reduce__
+    """
+    temp_dir = _get_temp_folder()
+    b = BytesIO(packed_keras_model)
+    with zipfile.ZipFile(b, "r", zipfile.ZIP_DEFLATED) as zf:
+        for path in zf.namelist():
+            dest = os.path.join(temp_dir, path)
```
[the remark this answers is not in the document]
adriangb: Yep

scikeras/__init__.py
adriangb: Great point. This was just a blunder on my part, I'll just import it as `_saving_utils` and keep it private.

stsievert: I think you can remove the import line. IIRC, it doesn't change this usage:

```
import scikeras._saving_utils
from scikeras import _saving_utils
```

scikeras/_saving_utils.py
adriangb: I may be missing something, so sorry if this causes confusion:

> Why can't _create_all_weights be a regular function?

I'm not sure what you mean by "a regular function"? `Optimizer._create_all_weights` is a function, `_temp_create_all_weights` is a function that replaces `Optimizer._create_all_weights` so that weights get restored after `Optimizer._create_all_weights` is called for the first time, and that then cleans up the extra weights and un monkey patches itself.

> It'd be a little cleaner if this all moved to the Optimizer class

We can't directly edit the `Optimizer` class, only monkey patch methods onto it. I'm not sure what the advantage of monkey patching `_restore_weights` onto `Optimizer` is vs. referencing it directly.

We also need to call `_restore_optimizer_weights` from `unpack_keras_model` because the implementation is not recursive: when optimizers are serialized as part of a model, `Optimizer.__reduce__` is never called.

scikeras/_saving_utils.py
adriangb: They are very simple, yes, but still required to have a single function we can attach as `__reduce__`. I guess one alternative, that also applies to the other monkey patching, would be to do:

```python
from tf.keras import metrics

class Metric(metrics.Metric):
    def __reduce__(...):
       ...
metrics.Metric = Metric
```

I guess this could work, but I'm not sure that it's cleaner.

scikeras/__init__.py
adriangb: I'm planning on doing:
```python3
from scikeras import _saving_utils
keras.Model.__reduce__ = _saving_utils.pack_keras_model
```
I think that's what you're suggesting but just checking

scikeras/_saving_utils.py
stsievert: Does that example change the user's version of `tf.keras.Metric`? Should a user ever import it? Or is it only used in SciKeras?

If a user passes a custom metric, can that metric be serialized?

adriangb: We need to change the global version of `Metric` in order for all metrics (including custom user metrics) to be serializable. I don't see any downside to this as it doesn't break anything as far as I can tell.

The alternative to modifying the global/user version would be to have a hook in `__init__` much like we have now that monkey patches the method onto live instances of `Model`, `Optimizer`, etc. instead of onto the classes like I am doing here. The con of this is that it (1) requires more code and (2) means that if a user were to define a custom parameter and passed an `Optimizer` to it (something like `KerasClassifier(..., my_optimizers=[Adam(), SGD()])` they'd get some cryptic error about `RLock` objects or they might not even get an error and instead get unpredictable results

Issue #124, opened 2020-11-05
DOC: Add a page to help users transition from tf.keras.wrappers.scikit_learn


adriangb:
We should have a page noting the differences (not features) vs. the old wrappers. I think so far we only have:

- `fit`/`predict` arguments should be passed to the constructor instead of as kwargs to `fit`/`predict`.
- Renamed `build_fn` to `model`
- No longer need to declare defaults in `build_fn`, instead they should be declared in the SciKeras constructor
- To get automatic one-hot encoding with `categorical_crossentropy` `loss="categorical_crossentropy"` (or `loss=CategoricalCrossentropy`, etc.) needs to be passed to the SciKeras constructor (even for user-compiled models) since we no longer introspect compiled models.

stsievert:
I thought SciKeras aimed to be compatible with `tf.keras.wrappers.scikit_learn`?

adriangb:
Yes, these are minor changes. I think all of these we've discussed together previously. An example with all of these changes (copied from #112):

```python3
def build_model(optimizer='adam'):
    model = Sequential()
    model.add(Flatten(input_shape=(28, 28)))
    model.add(Dense(512, activation='relu'))
    model.add(Dropout(0.2))
    model.add(Dense(10, activation='softmax'))
    model.compile(optimizer=optimizer,
              loss='categorical_crossentropy',
              metrics=['accuracy'])

    return model
    
model = KerasClassifier(build_fn=build_model)
```

New:
```python3
def build_model(optimizer):
    model = Sequential()
    model.add(Flatten(input_shape=(28, 28)))
    model.add(Dense(512, activation='relu'))
    model.add(Dropout(0.2))
    model.add(Dense(10, activation='softmax'))
    model.compile(optimizer=optimizer,
              loss='categorical_crossentropy',
              metrics=['accuracy'])

    return model
    
model = KerasClassifier(model=build_model, optimizer="rmsprop", loss="categorical_crossentropy")
```

stsievert:
> To get automatic one-hot encoding with categorical_crossentropy loss="categorical_crossentropy" (or loss=CategoricalCrossentropy, etc.) needs to be passed to the SciKeras constructor (even for user-compiled models) since we no longer introspect compiled models.

Why not specify a default of `loss="categorical_crossentropy"` for KerasClassifier? That'd also resolve #112, correct? Does it have any side effects?

> * fit/predict arguments should be passed to the constructor instead of as kwargs to fit/predict.
[...]

PR #133, opened 2020-11-25
DOC: add comparison to TF wrappers

### LINKED ISSUE #132: DOC: Add a doc page detailing advantages (and disadvantages?) vs. TF skelarn wrappers
A basic question that users might ask before even trying SciKeras or reading the docs is: _why should I use SciKeras over tf.keras.wrappers.scikit_learn_?

To answer this, I think we should put 5-6 bullet points on our README.md as well as docs.

adriangb:
Closes #132


README.md
```diff
@@ -6,6 +6,20 @@
 
 Scikit-Learn compatible wrappers for Keras Models.
 
+## Why SciKeras?
+
+SciKeras is derived from and API compatible with `tf.keras.wrappers.scikit_learn`. The original TensorFlow (TF) wrappers are not actively mantained, and [may be deprecated](https://github.com/tensorflow/tensorflow/pull/37201#pullrequestreview-391650001) at some point. In addition, they have many incompatibilities with both the Keras ecosystem and the Scikit-Learn ecosystem. SciKeras attempts to resolve these issues by providing mantained, documented wrappers that are fully compatible with the entire Scikit-Learn and Keras ecosystems. Some advantages over the TF wrappers are:
```
stsievert, line 11: > Some advantages over the TF wrappers are:

Is there documentation on how to go from TF wrappers to SciKeras wrappers?

adriangb: Not currently. That would also be a good idea to add here. This stuff should also get added to the RTD docs.

stsievert: > That would also be a good idea to add here

When I visted the RTD documentation, I expected to see the transition guide  at the top of the "getting started" page.

> This stuff should also get added to the RTD docs.

Yes, I agree. I tend to create READMEs that point directly to the HTML documentation. If I put anything on the README, it's very developer oriented (here's how to clone, directory structure, etc).

adriangb: I think the "why SciKeras" section should be in both so that people who discover the package via GitHub aren't confused as to why it even exists before they open the docs.

adriangb: I'm going to merge this as is (without any transition guide work). We have #124 to track the transition guide, and in particular, I would like to implement your suggestion in https://github.com/adriangb/scikeras/issues/124#issuecomment-723143535 since that will reduce the required changes. These changes would go beyond the scope of _this_ PR.

PR #138, opened 2020-11-27
DOC: add transition guide & ease fit/predict kwarg compatibility

### LINKED ISSUE #124: DOC: Add a page to help users transition from tf.keras.wrappers.scikit_learn

adriangb:
Closes #124


README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+proof your code, you should instead declare these parameters in your constructor:
+
+```python
+clf = KerasClassifier(batch_size=32)
+```
+
+Or to declare separate values for `fit` and `predict`:
+
+```python
+clf = KerasClassifier(fit__batch_size=32, predict__batch_size=32)
```
stsievert, line 69: Style nit: could "predict__batch_size" be larger, maybe 1,000 or 10,000?

README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+```
+
+### Renaming of `build_fn` to `model`
+
+SciKeras renamed the constructor argument `build_fn` to `model`. In a future release,
+passing `build_fn` as a _keyword_ argument will raise a `TypeError`. Passing it as a positional
+argument remains unchanged. You can make the following change to future proof your code:
+
+```python
+clf = KerasClassifier(build_fn=...)  # from
```
stsievert, line 79: Style not: maybe it'd be better to have these changes with diff highlighting?

README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+clf = KerasClassifier(model=...)  # to
+```
+
+### Default arguments in `build_fn`/`model`
+
+SciKeras will no longer introspect your callable `model` for _user defined_ parameters. You
+must now "declare" them as keyword arguments to the constructor if you want them to be
+tunable parameters (i.e. settable via `set_params`):
```
stsievert, line 87: Did Keras ever introspect?

On master, doesn't parameter routing remove the need for declaration?

README.md
adriangb: Yep that makes sense, will change it.

README.md
adriangb: I didn't know that's possible! That is obviously a better solution.

README.md
adriangb: > Did Keras ever introspect?

Yes, in Keras the following works:

```python3
def get_model(my_param=123):
    ...
clf = KerasClassifier(get_model)
clf.set_params(my_param=321)
```

In SciKeras, this would not work. You would have to use `KerasClassifier(get_model, my_param=...)`.
But if `set_params` is not being called, no changes are needed.

> On master, doesn't parameter routing remove the need for declaration?

I didn't understand the question, could you rephrase? Sorry...

README.md
adriangb: This also worked in the RST docs. Thanks for the suggestion.

scikeras/wrappers.py
```diff
@@ -712,13 +712,30 @@ def fit(self, X, y, sample_weight=None) -> "BaseWrapper":
...
+            warnings.warn(
+                "``kwargs`` will be removed in a future release of SciKeras."
+                f"Instead, set fit arguments at initialization (i.e., ``BaseWrapper({k}={v})``)"
+            )
+        self.set_params(
+            **{
+                (k if k.startswith("fit__") else "fit__" + k): v
+                for k, v in kwargs.items()
+            }
+        )
```
adriangb, line 737: @stsievert  what are your thoughts on these implementations? The main con I see is that subsequent calls will re-use the previous parameters. That is:

```python
est.fit(..., batch_size=32)
...
est.fit(...)  # implicitly uses batch_size=32 with no warnings to the user
```

stsievert: :-1: on this implementation. As a user, I would expect the params to behave the same unless I change them. Let the user do that, and issue an appropriate warning (like is being done).

[PEP 20][1]: "Explicit is better than implicit."

[1]:https://www.python.org/dev/peps/pep-0020/#the-zen-of-python

README.md
stsievert: > I didn't understand the question, could you rephrase? Sorry...

I thought SciKeras supported prefixing `model__` to various parameters to route to the model building function. That is, I though this worked:

``` python
def get_model(foo=123):
    ...
clf = KerasClassifier(get_model)
clf.set_params(model__foo=456)
```

I thought SciKeras had deprecated setting parameters without a `model__` prefix (e.g., `clf.set_params(foo=456)`)... or did we walk that back?

adriangb: It does, but it's not _required_ unless there is a collision (e.g. your model building function for whatever reason has a parameter called `batch_size`). And since the previous implementation does not support that, I don't see why I should be mentioned in the transition guide.

scikeras/wrappers.py
adriangb: Which part of the implementation are you 👎 on? This is basically what you proposed, plus the prefixing of `fit__` and `predict__`. Adding this prefix makes it behave more like the old implement, in the sense that passing a kwarg to `fit` won't "set" it for `predict` and visa versa.

Overall, I think there are 3 options:
1) We don't support `**kwargs` at all and _force_ users to make the change.
2) This implementation.
3) This implementation + we "unset"/"revert" the kwargs at the end. I haven't tested implementing this, and I do think it would add some complexity, but in theory it should be possible.

adriangb: `wrappers.py` diff cleaned up. (3) is looking pretty good now.

scikeras/wrappers.py
```diff
@@ -697,26 +697,41 @@ def feature_encoder(self):
...
+        for k, v in kwargs.items():
+            warnings.warn(
+                "``kwargs`` will be removed in a future release of SciKeras."
+                f"Instead, set fit arguments at initialization (i.e., ``BaseWrapper({k}={v})``)"
+            )
+        kwargs = {
+            (k if k.startswith("fit__") else "fit__" + k): v for k, v in kwargs.items()
+        }
+        existing_kwargs = {k: v for k, v in self.get_params().items() if k in kwargs}
+        self.set_params(**kwargs)
```
stsievert, line 733: :-1:

I'm still not seeing the need to set parameters. It looks like `set_params` is a vehicle to get these `kwargs` to `self._fit` or `self._fit_keras_model`. Why can't `kwargs` be passed to those functions and this processing be done there?

scikeras/wrappers.py
```diff
@@ -878,26 +893,42 @@ def partial_fit(self, X, y, sample_weight=None) -> "BaseWrapper":
...
         """Returns predictions for the given test data.
 
         Parameters
         ----------
         X : Union[array-like, sparse matrix, dataframe] of shape (n_samples, n_features)
             Training samples where n_samples is the number of samples
             and n_features is the number of features.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to ``Model.predict``.
+            This paremeter will be removed in a future release of SciKeras,
```
stsievert, line 906: Why are you implementing these parameters and immediately saying this implementation will be removed in a future release?

scikeras/wrappers.py
adriangb: You were the one who originally suggested this approach in https://github.com/adriangb/scikeras/issues/124#issuecomment-723143535

We could pass `**kwargs` around. Perhaps at this level of complexity/LOC that is a better approach. I'll try implementing it that way instead and see how it turns out.

scikeras/wrappers.py
adriangb: This is only to ease backward compatibility. We do _not_ want this `**kwargs` interface permanently. The idea would be to implement it until a `1.0.0` release or something, to help users from the old wrappers transition. Then we'd remove this interface. The alternative is to make it a required breaking change to transition to SciKeras.

scikeras/wrappers.py
adriangb: I was able to implement this passing `**kwargs` around directly. It is indeed simpler and cleaner.

scikeras/wrappers.py
```diff
@@ -697,34 +701,46 @@ def feature_encoder(self):
...
             True labels for X.
         sample_weight : array-like of shape (n_samples,), default=None
             Array of weights that are assigned to individual samples.
             If not provided, then each sample is given unit weight.
-
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to ``Model.fit``.
+            This paremeter will be removed in a future release of SciKeras,
+            instead set fit arguments at initialization
+            (e.g., ``BaseWrapper(epochs=10, ...)``)
```
stsievert, line 721:
```suggestion
            (e.g., ``BaseWrapper(epochs=10, fit__batch_size=32, ...)``)
```

scikeras/wrappers.py
```diff
@@ -878,26 +898,36 @@ def partial_fit(self, X, y, sample_weight=None) -> "BaseWrapper":
...
         Parameters
         ----------
         X : Union[array-like, sparse matrix, dataframe] of shape (n_samples, n_features)
             Training samples where n_samples is the number of samples
             and n_features is the number of features.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to ``Model.predict``.
+            This paremeter will be removed in a future release of SciKeras,
+            instead set fit arguments at initialization
+            (e.g., ``BaseWrapper(batch_size=10, ...)``)
```
stsievert, line 913:
```suggestion
            (e.g., ``BaseWrapper(verbose=1, predict__batch_size=1000, ...)``)
```

scikeras/wrappers.py
stsievert: Good idea.

Style nit: I might add this text:

> **kwargs : Dict[str, Any]
Extra arguments to route to ``Model.predict``.
>
> *This functionality has been deprecated, and will be removed in SciKeras 1.0.0. These parameters can also be specified by prefixing `fit__` to a parameter at initialization; e.g, `BaseWrapper(..., fit__batch_size=32, predict__batch_size=1000)`.

I'm not sure I would deprecate support for the Keras APIs.

adriangb: > I'm not sure I would deprecate support for the Keras APIs.

I think it's generally important to have only one obvious way to do things. And in sklearn, parameters like `shuffle` and `validation_frac` are set in `__init__`, not `fit`. So that's how I think we should do it in SciKeras.

adriangb: How about the following: we change the wording from _will be deprecated_ to _is scheduled for deprecation_ and link to an issue that we can keep open for discussion. If enough good use cases come up, we can "reverse course" by simply deleting the warning in a future release. Is this a good compromise @stsievert ?

stsievert: I suggest keeping the warning, and changing the wording to "Deprecated in version v0.2.x": a behavior is deprecated if  a warning is issued by definition.

> If enough good use cases come up, we can "reverse course" by simply deleting the warning in a future release. 

Deprecations are not final. All "deprecate" means is "express disapproval of" – we can stop expressing disapproval or delete the warning if a good use case comes along.

> I think it's generally important to have only one obvious way to do things.

Touche. As a corollary, that means the error raised should suggest a resolution method.

adriangb: Sounds good, thanks for the input. I edited the warning to:
```
**kwargs has been deprecated, and will be removed in SciKeras 1.0.0.
Instead, set predict arguments at initialization (i.e., BaseWrapper({k}={v}) )
```
Where k and v are the kwarg name/key and value respectively. I think this is in line with what you have suggested.

scikeras/wrappers.py
```diff
@@ -697,34 +701,46 @@ def feature_encoder(self):
...
+            e.g, `BaseWrapper(..., fit__batch_size=32, predict__batch_size=1000)`.
         Returns
         -------
         BaseWrapper
             A reference to the instance that can be chain called
             (ex: instance.fit(X,y).transform(X) )
         """
+        for k, v in kwargs.items():
+            warnings.warn(
+                "``**kwargs`` has been deprecated in SciKeras 0.2.1, and will be 1.0.0."
```
stsievert, line 730:
```suggestion
                "``**kwargs`` has been deprecated in SciKeras 0.2.1 and support will be removed be 1.0.0."
```

This goes for several other places too.

scikeras/wrappers.py
```diff
@@ -878,26 +898,36 @@ def partial_fit(self, X, y, sample_weight=None) -> "BaseWrapper":
...
+            e.g, `BaseWrapper(..., fit__batch_size=32, predict__batch_size=1000)`.
 
         Returns
         -------
         array-like
             Predictions, of shape shape (n_samples,) or (n_samples, n_outputs).
         """
+        for k, v in kwargs.items():
+            warnings.warn(
+                "``**kwargs`` has been deprecated in SciKeras 0.2.1, and will be 1.0.0."
```
stsievert, line 922:
```suggestion
                "``**kwargs`` has been deprecated in SciKeras 0.2.1 and support will be removed be 1.0.0."
```

scikeras/wrappers.py
```diff
@@ -1394,6 +1436,12 @@ def predict_proba(self, X):
             SciKeras will return an array of shape (n_samples, 2)
             (instead of `(n_sample, 1)` as in Keras).
         """
+        for k, v in kwargs.items():
+            warnings.warn(
+                "``**kwargs`` has been deprecated in SciKeras 0.2.1, and will be 1.0.0."
```
stsievert, line 1441:
```suggestion
                "``**kwargs`` has been deprecated in SciKeras 0.2.1 and support will be removed be 1.0.0."
```

scikeras/wrappers.py
adriangb: Thanks, I replaced it in 7dd1cd4.

adriangb:
@stsievert are we ready to merge this? In light of https://github.com/dask/dask-ml/issues/764#issuecomment-736895951 I'd like to cut a patch release, and I think it would be good if this is in it so that it can be a decent long term release that Keras users can switch over into.

README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+SciKeras does however have some backward incompatible changes:
+
+### Automatic one-hot encoding of targets for categorical crossentropy losses
+
+SciKeras will not longer implicitly inspect your Model's loss function to determine if
+it needs to one-hot encode your target to match the loss function. Instead, you must explicitly
+pass your loss function to the constructor:
+
+```python
+clf = KerasClassifier(loss="categorical_crossentropy")  # or loss=CategoricalCrossentropy(), etc.
```
stsievert, line 54:
```suggestion
clf = KerasClassifier(loss="categorical_crossentropy")
```

README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+### Removal of `**kwargs` from fit and predict
+
+In a future release of SciKeras, `**kwargs` will be removed from fit and predict. To future
+proof your code, you should instead declare these parameters in your constructor:
+
+```diff
+- clf = KerasClassifier(...)
++ clf = KerasClassifier(.., batch_size=32)
+- clf.fit(..., batch_size=32)
++ clf.fit(...)
```
stsievert, line 66: Style nit:

``` diff
- clf = KerasClassifier(...)
- clf.fit(..., batch_size=32)
+ clf = KerasClassifier(.., batch_size=32)
+ clf.fit(...)
```

README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+```
+
+Or to declare separate values for `fit` and `predict`:
+
+```python
+clf = KerasClassifier(fit__batch_size=32, predict__batch_size=10000)
```
stsievert, line 72: Tangential to this PR, I think these would be good defaults.

README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+### Default arguments in `build_fn`/`model`
+
+SciKeras will no longer introspect your callable `model` for _user defined_ parameters. You
+must now "declare" them as keyword arguments to the constructor if you want them to be
+tunable parameters (i.e. settable via `set_params`):
+
+```diff
+- def get_model(my_param=123):
++ def get_model(my_param):  # You can optionally remove the default here
+    ...
```
stsievert, line 95:
```suggestion
    ...
    return model
```

README.md
```diff
@@ -32,6 +33,73 @@ pip install scikeras
...
+- clf = KerasClassifier(get_model)
++ clf = KerasClassifier(get_model, my_param=123)
```
stsievert, line 97:
```suggestion
+ clf = KerasClassifier(get_model, my_param=123)  # option 1
+ clf = KerasClassifier(get_model, model__my_param=123)  # option 2
```

README.md
```diff
@@ -15,6 +15,7 @@ SciKeras attempts to resolve these issues by providing maintained, documented wr
 entire Scikit-Learn and Keras ecosystems. Some advantages over the TF wrappers are:
```
stsievert, line 15: I still suggest this outline for README.md:

> **Why should I use SciKeras?**
> The motivation is at https://scikeras.readthedocs.io/
>
> **How do I transition from TF?**
> See https://scikeras.readthedocs.io/en/latest/quickstart.html
>
> **Basic development workflow**
> 1. ...
> 2. ...

I think the only people who read READMEs are people browsing GitHub (so links don't bother them), and developers (who aren't interested in these questions anyways).

stsievert: I commented README.md because it showed up first in GitHub's diff, but it looks like my comments also apply to index.rst.

docs/source/index.rst
```diff
@@ -29,6 +19,92 @@ Overall, SciKeras aims at being as flexible as Keras while having a clean interf
...
+
+SciKeras will not longer implicitly inspect your Model's loss function to determine if
+it needs to one-hot encode your target to match the loss function. Instead, you must explicitly
+pass your loss function to the constructor:
+
+.. code:: python
+
+   clf = KerasClassifier(loss="categorical_crossentropy")  # or loss=CategoricalCrossentropy(), etc.
+
+Removal of ``**kwargs`` from fit and predict
```
stsievert, line 59:
```suggestion
Removal of keyword arguments from fit and predict
```

docs/source/index.rst
```diff
@@ -29,6 +19,92 @@ Overall, SciKeras aims at being as flexible as Keras while having a clean interf
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In a future release of SciKeras, ``**kwargs`` will be removed from fit and predict. To future
```
stsievert, line 62:
```suggestion
In a future release of SciKeras, keyword arguments will be removed from fit and predict. To future
```

docs/source/index.rst
```diff
@@ -29,6 +19,92 @@ Overall, SciKeras aims at being as flexible as Keras while having a clean interf
...
+   - clf.fit(..., batch_size=32)
+   + clf.fit(...)
+
+Or to declare separate values for ``fit`` and ``predict``:
+
+.. code:: python
+
+   clf = KerasClassifier(fit__batch_size=32, predict__batch_size=10000)
+
+Renaming of ``build_fn`` to ``model``
```
stsievert, line 78: Are these warnings in `index.rst` necessary? I'm having a hard time seeing the reason to feature the following on the front page of the documentation:

* Removal of keyword args
* Renaming `build_fn` to `model`
* Model introspection

All of those issue warnings with tips on how to resolve. Why are they on the front page of the documentation? When I write docs, the front page always describes why a user should use my software, not issues they might face.

I might dedicate a separate page to potential hotspots when transitioning from TF to SciKeras.

scikeras/wrappers.py
```diff
@@ -463,6 +464,8 @@ def _fit_keras_model(
             Number of epochs for which the model will be trained.
         initial_epoch : int
             Epoch at which to begin training.
+        **kwargs : Dict[str, Any]
+            Extra arguments to route to ``Model.fit``.
```
stsievert, line 468: What arguments can these be? I would include a link to the Keras API docs here. https://keras.io/api/models/model_training_apis/

scikeras/wrappers.py
```diff
@@ -697,34 +701,46 @@ def feature_encoder(self):
...
         Returns
         -------
         BaseWrapper
             A reference to the instance that can be chain called
             (ex: instance.fit(X,y).transform(X) )
         """
+        for k, v in kwargs.items():
+            warnings.warn(
+                "``**kwargs`` has been deprecated in SciKeras 0.2.1 and support will be removed be 1.0.0."
+                f" Instead, set fit arguments at initialization (i.e., ``BaseWrapper({k}={v})``)"
```
stsievert, line 731:
```suggestion
                f" Instead, set fit arguments at initialization (e.g, ``BaseWrapper({k}={v})`` "
                f"or ``BaseWrapper(fit__{k}={v})``)"
```

scikeras/wrappers.py
```diff
@@ -878,26 +898,36 @@ def partial_fit(self, X, y, sample_weight=None) -> "BaseWrapper":
...
 
         Returns
         -------
         array-like
             Predictions, of shape shape (n_samples,) or (n_samples, n_outputs).
         """
+        for k, v in kwargs.items():
+            warnings.warn(
+                "``**kwargs`` has been deprecated in SciKeras 0.2.1 and support will be removed be 1.0.0."
+                f" Instead, set predict arguments at initialization (i.e., ``BaseWrapper({k}={v})``)"
```
stsievert, line 923:
```suggestion
                f" Instead, set predict arguments at initialization (i.e., ``BaseWrapper({k}={v})`` "
                f"or ``BaseWrapper(predict__{k}={v})``)"
```

scikeras/wrappers.py
```diff
@@ -1394,6 +1436,12 @@ def predict_proba(self, X):
             SciKeras will return an array of shape (n_samples, 2)
             (instead of `(n_sample, 1)` as in Keras).
         """
+        for k, v in kwargs.items():
+            warnings.warn(
+                "``**kwargs`` has been deprecated in SciKeras 0.2.1 and support will be removed be 1.0.0."
+                f" Instead, set predict_proba arguments at initialization (i.e., ``BaseWrapper({k}={v})``)"
```
stsievert, line 1442:
```suggestion
                f" Instead, set predict_proba arguments at initialization (i.e, ``BaseWrapper({k}={v})`` "
                f"or ``BaseWrapper(predict_proba__{k}={v})``"
```

Does that behavior work?

docs/source/index.rst
adriangb: Yeah a separate page should work and sounds good. Thanks.

scikeras/wrappers.py
adriangb: The current docs are at https://www.tensorflow.org/api_docs/python/tf/keras/Model?version=nightly#fit but yeah, I'll include that.

scikeras/wrappers.py
adriangb: No, for the purposes of parameter routing `predict_proba` uses `predict__` since they both call `predict__` in Keras. We could alias them if you think that might make things easier. It might also be good to have an error or at least a warning for unknown prefixes. All of this would have to be a separate PR.

docs/source/index.rst
adriangb: Is this wording accurate? We don't plan on removing keyword arguments in the sense that we would only user or allow positional arguments. Maybe something like `Removal of arbitrary keyword arguments passed via **kwargs`? But that's pretty long, especially for a title...

README.md
adriangb: Got it. I'll mirror changes in README.md to index.rst (or a new page, as discussed).

docs/source/index.rst
stsievert: Why not "Removal of Keras keyword arguments passed to fit/predict"?

scikeras/wrappers.py
stsievert: I've edited the suggested change to use the `predict__` prefix.

docs/source/index.rst
adriangb: I'm going to go with `Variable keyword arguments in fit and predict` as the title.

README.md
adriangb: Added

adriangb:
@stsievert I incorporated your suggestions manually instead of committing them because I moved all of this content to a `migrations.rst` page instead of having it in `README.md` and `index.rst`. Otherwise I think I largely incorporated it all of the suggestions.

PR #143, opened 2020-11-29
ENH: Check input dimensions against the initialized model_

### LINKED ISSUE #106: Check number of inputs to Keras model
Currently, the number of outputs of the Keras model is checked:

https://github.com/adriangb/scikeras/blob/9f633f5c21a45b43986a8d779c3ef9e3005038bb/scikeras/wrappers.py#L418-L419

It'd be nice if the number of inputs could be checked too I think. This should be possible according to the docs, which says a Keras Model has an `.inputs` attribute: https://keras.io/api/models/model/

adriangb:
Closes #106


adriangb:
@stsievert tagging you to take a look whenever you have a chance. This is addressing an issue you had opened (#106)

scikeras/wrappers.py
```diff
@@ -526,22 +526,46 @@ def _fit_keras_model(
...
+        # hence we cannot just check y here, we need the user (or the
+        # data transformer) to tell us what to expect via n_outputs_expected_
+        # n_outputs_expected_ is generated by data transformers
+        # and recovered via target_encoder_.get_meta();
+        # we recognize it but do not force it to be
+        # generated to avoid forcing users to subclass
+        # generic transformers just to make them SciKeras compatible
+        n_out_expect = getattr(self, "n_outputs_expected_", None)
+        if n_out_expect:
+            if n_out_expect != len(self.model_.outputs):
```
stsievert, line 540: Style nit: I'd collapse this if-statement:

``` python
n_out_expect = getattr(self, "n_outputs_expected_", None)
if n_out_expect and n_out_expect != len(self.model_.outputs):
    ...
```

adriangb: Yep I'll implement that.

Do you think we should redefine `n_outputs_expected_: int` -> `expected_output_shape_: Union[List[int], Dict[str, int]]` so that each element/key can map to the expected output shape for each output?

stsievert: > Do you think we should redefine n_outputs_expected_: int -> expected_output_shape_: Union[List[int], Dict[str, int]] so that each element/key can map to the expected output shape for each output?

I don't think so. An integer suffices for basic use cases (regressors and classifiers). I can only see output shape being relevant for autoencoders.

adriangb: I'm thinking of multi-input or multi-output models.

stsievert: > I'm thinking of multi-input or multi-output models (as to why it would be a list/dict).

Maybe that's not a bad idea. I think the type of `n_outputs_expected` should be `Union[int, List[int], Dict[str, int]]`. That supports more complex models and simple user-defined transformers.

In this function that will mean something like this:

``` python
n_out = getattr(self, "n_outputs_expected", None)
if isinstance(n_out, int):
    n_out = [n_out]
if n_out and n_out != len(self.model_.outputs):
    ...
```

adriangb: Sounds good. I'll take a stab at implementing that the next couple of days.

adriangb: I see two possible issues with https://github.com/adriangb/scikeras/pull/143#discussion_r536928186:
1. This changes `n_outputs_expected_` for multiclass models. Previously, we set `n_outputs_expected_=1`. Now we'd set it to `n_outputs_expected_=n_classes_`, except for binary targets, where it is still `n_outputs_expected=1`.
2. This breaks the ability to use a 2 output units / softmax activation / categorical_crossentropy instead of the usual 1 output / sigmoid / binary_crossentropy. Since I don't think there's any use case for the former over the latter, I think this is OK to do, but I'd like your input on this @stsievert

Other than those two changes, I think this looks pretty good and should protect against several mistakes users can make.

adriangb: I'll put it in the docs along with this PR, but just to be clear, we are saying that `n_outputs_expected_` now contains information on both the number of outputs and the number of output units in each output:
* `1`: a single output, a single output unit
* `5`: a single output, 5 output units (5 classes or 5 regression targets)
* `[1]`: a single output, a single output unit. Equivalent a bare `1`
* `[1, 1]`: two output, each with a single output unit
* `{"Out1": 1", ...}`: the output named "Out1" has a single output unit.

stsievert: > This breaks the ability to use a 2 output units / softmax activation / categorical_crossentropy instead of the usual 1 output / sigmoid / binary_crossentropy.

Can an error be raised? If I'm reading it correctly, the code would look like this:

``` python
if len_model_output == 2 and loss == "categorical_crossentropy":
    raise ValueError("... use binary_crossentropy ...")
```

adriangb: Yes, great idea! I think we should be able to do that within the default transformer since we have the loss function.
We can't check the activation (I mean, we _could_ but I'm not sure we want to introspect that much? maybe another PR), but I'm not sure why anyone would be using a softmax activation with binary_crossentropy loss.

stsievert: The activation isn't super important, it only influences the model (not the data flow). I might add it as a suggestion in the error message. Does Keras have a doc page that describes the conversion?

It looks like there's some relationship between `n_outputs == 2 and loss == "categorical_crossentropy"` and `n_outputs == 1 and loss == "binary_crossentropy"`. Why shouldn't relationship be mentioned in the error message?

adriangb: Yes, we can/should mention the relationship.

I would put the following in `transformers.py::ClassifierLabelEncoder.fit`:
```python
if target_type_ == "binary" and loss == "categorical_crossentropy":
    raise ValueError(
        'SciKeras does not support use of binary targets with "categorical_crossentropy". Instead, use'
        ' "binary_crossentropy". To change this behavior, override the default transformer (see link_to_docs).'
    )
```

The number of outputs are then _set_ here, and compared later in `wrappers::BaseWrapper. _check_model_outputs` (once we have the built model).

How does this sound?

stsievert: How 'bout the error message below? It's not clear what I would change in a custom transformer to clear this error.

> SciKeras does not support use of binary targets with `loss="categorical_crossentropy"`, or equivalently, SciKeras does not support binary targets with two output neurons. The following will resolve this error:
>
> * (recommended) Use a single output unit for binary targets w/ `loss="binary_crossentropy"`
> * Passing a custom transformer that does not have an attribute `n_outputs_expected_`.

How could I change `build_fn` to work for both `target_type=="binary"` and `target_type=="multiclass"`?

Would this error be raised if a compiled model is returned from `build_fn`?

adriangb: Maybe instead of checking the loss, we can do the real check in `_check_model_outputs`
```python3
if expected == 1 and actual == [2] and is stance(self.target_encoder_, ClassifierLabelEncoder):
    raise ValueError("ClassifierLabelEncoder does not support...")
```
I don't particularly like `isinstance` checks, but I do think this resolves ambiguity.

adriangb: Another alternative is of course to revert to https://github.com/adriangb/scikeras/pull/143/commits/c7e6723f8ff2536324b556939c19d955a2a0545a and try to continues this conversation in it's own issue/PR.

stsievert: I think I like checking the model better. But why is `n_outputs_expected_` necessary?

> Another alternative is of course to revert to c7e6723

Do you mean delete the release from PyPI/GitHub? If I were handling releases, I think I'd rather delay any checks on `n_outputs_expected_` to a later release.

adriangb: It's not _required_. There's other (perhaps better) ways to achieve the same thing.

That commit is in this branch. I was suggesting we revert to that commit in this branch and table this discussion for another issue/PR since the rest of the changes here stand on their own.

stsievert: > That commit is in this branch. I was suggesting we revert to that commit in this branch and table this discussion for another issue/PR since the rest of the changes here stand on their own.

Oh, I see. I think reverting and a separate issue is a good idea.

> It's not required. There's other (perhaps better) ways to achieve the same thing.

I think those (perhaps better) ideas are worth exploring.

adriangb: Sounds good. I reverted to c7e6723f8ff2536324b556939c19d955a2a0545a and implemented the collapsed logic in https://github.com/adriangb/scikeras/pull/143#discussion_r532327729. I'll resolve this thread and we can continue the discussion on the rest of the PR. I also opened #148 to track the discussion surrounding output validation. Thank you for the feedback thus far!

PR #155, opened 2020-12-28
TST: compare to sklearn.neural_network

### LINKED ISSUE #119: TST: Add comparisons to MLPClassifier/MLPRegressor in terms of output shapes and dtypes
Add more tests using sklearn classifiers/regressors that check output shapes and dtypes for different types of target types and target dtypes. Similar to the current [`test_multilabel_classification`](https://github.com/adriangb/scikeras/blob/9f633f5c21a45b43986a8d779c3ef9e3005038bb/tests/test_input_outputs.py#L177) test.

adriangb:
Closes #119.

An overhaul of `test_input_outputs.py` to combine tests and compare to `MLPRegressor`/`MLPClassifier`.

A couple of interesting notes:
- sklearn always returns np.int64 for multilabel-indicator tasks. This makes some sense, since the data is always `0/1`. But there is something to be said for returning the same dtype as the input.
- sklearn regressors always return np.float64, even if the input is int or float32. Returning floats even if the input is int makes sense, but it probably doesn't make sense to return higher precision floating points than the input, especially if Keras' backend dtype is usually float32. We might want to diverge from sklearn here and just return the raw data from Keras.


scikeras/utils/transformers.py
```diff
@@ -368,7 +368,7 @@ def inverse_transform(self, y: np.ndarray) -> np.ndarray:
             Keras Model predictions cast to the dtype and shape of the input
             targets.
         """
-        if self._y_dtype == np.float64 and y.dtype == np.float32:
+        if y.dtype == np.float32:
```
stsievert, line 371:
```suggestion
        if y.dtype.name == "float32":
```

.github/workflows/tests.yaml
```diff
@@ -103,7 +103,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [2.2.0]
+        tf-version: [2.4.0]
```
adriangb, line 106: These version bumps are needed in order to be able to get rid of `_windows_upcast_ints` and have the same tests run on all platforms (otherwise there would've been a mess of `pytest.skipif`).

This also coincides with the same version bump in #128 , so I think it makes sense to do. We could perhaps do the version bump in a separate PR from those two to keep the diffs saner.

stsievert: What will fail when this PR is run with TF v2.2? Will Windows users notice a difference?

adriangb: `test_input_dtype_conversion` will fail because on windows `_windows_upcast_ints` converts all ints to int64 before passing them to Keras.

This could be worked around by checking the dtype at a point before that (but the very fact that might work is an argument for _not_ doing that) or by using `pytest.skipif(os.platform == "nt")` or something.

Windows users won't notice any difference (I doubt the type conversion mattered in terms of performance).

I'm going to make a `develop` branch where I'll do the version bump separately and then redirect this PR, #128 and #143 to `develop`.

stsievert: So the only difference a Windows user will notice with TF 2.2 is that the input and output dtypes aren't exactly the same? They'll input `int32` and get out `int64` or vice versa?

Is that fixed under TF 2.4?

adriangb: Previously, when users/the data transformer handed data of dtype `int32` we transformed it to `int64` since TF crashed with `int32`. Since we never returned data to the users, they did not see the difference. That seems to be fixed in 2.4.0, so we can remove the extra conversion without users noticing any difference.

stsievert: Got it. If running this code with TF < 2.4.0/without `_windows_upcast_ints` will crash, I'd require TF >= 2.4.0.

I'd also keep the test that verifies `_windows_upcast_ints` works, even if it works with TF >= 2.4.0 (I can't tell if that test is changed).

adriangb: > Got it. If running this code with TF < 2.4.0/without _windows_upcast_ints will crash, I'd require TF >= 2.4.0.

Yep.

> I'd also keep the test that verifies _windows_upcast_ints works

I'm not sure I'm understanding. Why do we need a test to check if `_windows_upcast_ints` works if we're deleting the function?

stsievert: Why would you delete a test if it always passes?

What happens if TF regresses on this bug fix?

Tests are only written to ensure that the software works as expected for users. Did the test ensure developers can trust internal SciKeras code? If so, please delete it.

stsievert: > Why not delete a test if the [private function] has been removed? (paraphrased)

Why do you write tests? I write tests to ensure the software works as expected for the end user when they pick it up, and in every corner case too.

There's an argument for testing private behavior, or unit testing. But that's only appropriate for behemoth projects with many developers.

One of corner case is Windows for SciKeras. The fact that code has been deleted is irrelevant to test deletion. Again, why do you write tests? To ensure SciKeras works well for Windows users.

adriangb: I do realize the test diff is pretty large, and not particularly easy to parse. I am happy to try to rework this into multiple PRs, write out the new tests in pseudcode or anything else that might make review easier, just let me know what you think might help.

Regarding the question about test deletion, thank you for answering it even though it was a pretty poor question on my part. I think we both agree that writing unit tests for private functions like `_windows_upcast_ints` is not a good approach. Luckily, you already convinced me that that approach to testing is not optimal, so we never implemented `test_windows_upcast_ints`.

Because of your help and suggestions, I was able to remove `_windows_upcast_ints` in 1c1e17ba3f1afa795095205a9ac4d0638db12d69 (after bumping the TF version in 294183ae90c6bef926c9d8f2004c5380ea22bab7)  without having to change any tests. Thus all of the test changes in this PR are purely to re-organize our tests and implement comparisons to sklearn `MLP{Classifier,Regressor}`.

stsievert: > I do realize the test diff is pretty large, and not particularly easy to parse.

Don't worry about the diff. I only wrote that to let you know I only skimmed the tests. It sure sounds testing under control.

adriangb: Just checking, would you like me to leave this open for a more in-depth review of the tests, or should I move forward with this and/or #128, which are both internal refactors? No problem either way, just let me know.

PR #156, opened 2020-12-28
TMP: for windows testing

adriangb:


scikeras/utils/transformers.py
```diff
@@ -368,7 +368,7 @@ def inverse_transform(self, y: np.ndarray) -> np.ndarray:
             Keras Model predictions cast to the dtype and shape of the input
             targets.
         """
-        if self._y_dtype == np.float64 and y.dtype == np.float32:
+        if y.dtype == np.float32:
```
stsievert, line 371:
```suggestion
        if y.dtype.name == "float32":
```

adriangb: Glad to put the change into #155, but this is just a garbage branch I created because tests were failing on Windows.

I can't test locally on windows (I have MacOS), so I created this branch to avoid polluting your notifications and the commit log in #155 while I test fixes for Windows by pushing to GH.

scikeras/utils/transformers.py
```diff
@@ -368,7 +368,7 @@ def inverse_transform(self, y: np.ndarray) -> np.ndarray:
             Keras Model predictions cast to the dtype and shape of the input
             targets.
         """
-        if y.dtype == np.float32:
+        if y.dtype.name == "float32":
```
stsievert, line 371: :+1: this helps make SciKeras work with other array libraries (CuPy, Dask, XArray, ...).

adriangb: Great, thanks for the tip! I think I got most of them here since they were all in the file we're editing for this PR anyway, but I also opened https://github.com/adriangb/scikeras/issues/158 to try to track down any stray occurrences of this pattern in other files.

Issue #159, opened 2021-01-03
add batch_size = 'all' option?


john-veillette:
Hey devs, thanks for this package.

I see that **kwargs is being deprecated from .fit, which makes sense in the interest of making the API more consistent for grid search, etc. However, it still seems desirable to be able to specify batch_size based on the input dimensions in case users want to use vanilla rather than stochastic gradient descent. Maybe adding a batch_size = 'all' option in init would remove any need to specify batch_size in fit?

adriangb:
Hi 👋, glad you're finding it useful!

Would `all`=`X.shape[0]`?

john-veillette:
Yep! To replace the option to specify model.fit(X, y, batch_size = X.shape[0]) when fit(X, y, **kwargs) is removed, one could specify batch_size = 'all' on initialization. Should only take a few lines of code to implement.

adriangb:
Agreed. Will leave this issue open to track.

john-veillette:
Thanks!

stsievert:
Thats a decent argument for keeping fit `**kwargs`. I think compatibility with Keras is important.

adriangb:
Hey @john-veillette, I haven't implemented this yet because like @stsievert points out above, we may ended up keeping `**kwargs`.

That said, I do think there are some use cases where `**kwargs` doesn't cut it. The first that comes to mind is cross-val-score, where presumably the length of the dataset is changing but you have no control over the sizes / cannot pass `**kwargs.

For your use case, do you feel that you actually need `batch_size="all"` or could `**kwargs` work for you?

adriangb:
In addition to the implementation in #194, this could also be implemented via #167. Reference implementation here: https://www.adriangb.com/scikeras/refs/pull/167/merge/notebooks/DataTransformers.html#7.-Dynamically-setting-batch_size

john-veillette:
For my use case, **kwargs works just as well!

adriangb:
Implemented in #194

Issue #131, opened 2020-11-18
validation_data parameter to fit with preprocessing


gioxc88:
It's  pretty common in keras to pass `validation_data` to `fit` to monitor  how the loss behaves out of sample for each epoch.

I noticed that `scikeras` offers two solutions:

1. pass `validation_split` in the initialization 
2. pass `fit__validation_data` in the initialization 

Since I need to preprocess `X` and `y` before fitting I cannot use validation_split, because this would cause data leakage so I should opt for solution 2 but this makes the model stateful since the `fit__validation_data` is attached to the `model` instance.

Do you have any suggestion for this problem?

Many thanks

adriangb:
This is indeed an interesting question.

Could you do:
```python3
est.set_params(fit__validation_data=...)
est.fit(...)
est.set_params(fit__validation_data=None)
```
?

But perhaps we should put `validation_data` back into `fit`; I believe it's the only parameter that suffers from this complication / is actual data.

Could you elaborate on how you are preprocessing your data and how that would lead to data leakage? Are you using SciKeras' data pre/post processing interface?

gioxc88:
> This is indeed an interesting question.
> 
> Could you do:
> 
> ```python
> est.set_params(fit__validation_data=...)
> est.fit(...)
> est.set_params(fit__validation_data=None)
> ```
> 
> ?
> 
> But perhaps we should put `validation_data` back into `fit`; I believe it's the only parameter that suffers from this complication / is actual data.
> 
> Could you elaborate on how you are preprocessing your data and how that would lead to data leakage? Are you using SciKeras' data pre/post processing interface?

Thank you for the answer.
I am using `scikeras` as part of a bigger project and I always use `sklearn` like transformers to preprocess data inside a `pipeline`,  so basically `scikeras.KerasRegressor` instance would be last step of the `pipeline`.

Say for example (simple case) I want to use `StandardScaler` in the preprocessing step.
Passing `validation_split` means that I have a combined `[X_train, X_val]` dataset which contains both training and validation data. I apply preprocessing to the combined dataset and then pass it to `scikeras.KerasRegressor` which splits it according to `validation_split`. In this case both `X_train` and `X_val` are standardized using all data which causes leakage. 

Of course there are solutions to this, for example preprocess  `X_train` and `X_val` separeately, then combine them and figure out `validation_split = len(X_val) // (len(X_train) + len(X_val))`.

[...]

PR #167, opened 2021-01-16
ENH: add dependency injection point to transform X & y together

### LINKED ISSUE #160: Dealing with variable length inputs
Let's assume we are working with variable length inputs. One of the strongest parts in using `tf.data.Dataset` is the ability to pad batches as they come.

But since `scikit-learn`'s API is mainly focused around dataframes and arrays, incorporating this is kind of hard. Obviously, you can pad everything, but this can be a huge waste of memory. I'm trying to work with the `sklearn.pipeline.Pipeline` object, and I thought to myself "alright, I'll just create a custom transformer at the end of my pipeline just before the model, and make it return a `tf.data.Dataset` object to later plug in my model. But this is not possible since the `.transform` signature only accepts X and not y, while you'll need both to work with `tf.data.Dataset`.

So assume we have 4 features for each data point, and each has it's own sequence length, for example a datapoint might look like this:

```python3
sample_features = {'a': [1,2,3], 'b': [1,2,3,4,5], 'c': 1, 'd': [1,2]}
sample_label = 0
```

How will I be able to manage this kind of dataset under scikit learn + scikeras?

adriangb:
Resolves #160

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -696,6 +696,29 @@ class BaseWrapper(BaseEstimator):
         """
         return FunctionTransformer()
 
+    @property
+    def dataset_transformer(self):
+        """Retrieve a transformer to be applied jointly to X & y.
+
+        It MUST accept a 2 element tuple as it's single input argument
+        to `fit`, `transform` and `inverse_transform` and the
+        latter two MUST also output a two element tuple.
+        The second element, corresponding to `y`, can be None
+        if the first element (corresponding to `X`) is a Dataset.
+
+        Metadata will be collected from ``get_metadata`` if
+        the transformer implements that method.
+        Override this method to implement a custom data transformer
+        for entire dataset.
+
+        Returns
+        -------
+        dataset_transformer
+            Transformer implementing the sklearn transformer
+            interface.
+        """
+        return FunctionTransformer()
+
     def fit(self, X, y, sample_weight=None, **kwargs) -> "BaseWrapper":
         """Constructs a new model with ``model`` & fit the model to ``(X, y)``.
 
@@ -774,8 +797,11 @@ class BaseWrapper(BaseEstimator):
         target_metadata = getattr(self.target_encoder_, "get_metadata", dict)()
         vars(self).update(**target_metadata)
         self.feature_encoder_ = self.feature_encoder.fit(X)
-        feature_meta = getattr(self.feature_encoder, "get_metadata", dict)()
+        feature_meta = getattr(self.feature_encoder_, "get_metadata", dict)()
         vars(self).update(**feature_meta)
+        self.dataset_transformer_ = self.dataset_transformer.fit((X, y))
+        dataset_meta = getattr(self.dataset_transformer_, "get_metadata", dict)()
+        vars(self).update(**dataset_meta)
 
         self.model_ = self._build_keras_model()
 
@@ -846,10 +872,13 @@ class BaseWrapper(BaseEstimator):
         if sample_weight is not None:
             X, y, sample_weight = self._validate_sample_weight(X, y, sample_weight)
 
-        y = self.target_encoder_.transform(y)
         X = self.feature_encoder_.transform(X)
 
-        self._check_model_compatibility(y)
+        if y is not None:
+            y = self.target_encoder_.transform(y)
+            self._check_model_compatibility(y)
+
+        X, y = self.dataset_transformer_.transform((X, y))
 
         self._fit_keras_model(
             X,
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -2,6 +2,7 @@
 import pickle
 
 from typing import Any, Dict
+from unittest.mock import patch
 
 import numpy as np
 import pytest
@@ -17,7 +18,8 @@ from sklearn.ensemble import (
 from sklearn.exceptions import NotFittedError
 from sklearn.model_selection import GridSearchCV, RandomizedSearchCV
 from sklearn.pipeline import Pipeline
-from sklearn.preprocessing import StandardScaler
+from sklearn.preprocessing import FunctionTransformer, StandardScaler
+from tensorflow.data import Dataset
 from tensorflow.keras import losses as losses_module
 from tensorflow.keras import metrics as metrics_module
 from tensorflow.keras.layers import Conv2D, Dense, Flatten, Input
@@ -27,7 +29,7 @@ from tensorflow.python.keras import backend as K
 from tensorflow.python.keras.utils.generic_utils import register_keras_serializable
 from tensorflow.python.keras.utils.np_utils import to_categorical
 
-from scikeras.wrappers import BaseWrapper, KerasClassifier, KerasRegressor
+from scikeras.wrappers import KerasClassifier, KerasRegressor
 
 from .mlp_models import dynamic_classifier, dynamic_regressor
 from .testing_utils import basic_checks
@@ -812,3 +814,34 @@ class TestInitialize:
         np.testing.assert_allclose(y_pred_keras, y_pred_scikeras)
         # Check that we are still using the same model object
         assert est.model_ is m2
+
+
+class TestDatasetTransformer:
+    def test_conversion_to_dataset(self):
+        inp = Input((1,))
+        out = Dense(1, activation="sigmoid")(inp)
+        m = Model(inp, out)
+        m.compile(loss="bce")
+
+        class MyWrapper(KerasClassifier):
+            @property
+            def dataset_transformer(self):
+                f = lambda x_y: (Dataset.from_tensor_slices(x_y), None)
+                return FunctionTransformer(f)
+
+        est = MyWrapper(m)
+        X = np.random.random((100, 1))
+        y = np.array(["a", "b"] * 50, dtype=str)
+        fit_orig = m.fit
+
+        def check_fit(**kwargs):
+            assert isinstance(kwargs["x"], Dataset)
+            assert kwargs["y"] is None
+            return fit_orig(**kwargs)
+
+        with patch.object(m, "fit", new=check_fit):
+            est.fit(X, y)
+        y_pred = est.predict(X)
+        assert y_pred.dtype == y.dtype
+        assert y_pred.shape == y.shape
+        assert set(y_pred).issubset(set(y))
--- a/tests/test_param_routing.py
+++ b/tests/test_param_routing.py
@@ -23,6 +23,7 @@ keras_classifier_base_meta_set = {
     "n_outputs_",
     "feature_encoder_",
     "target_encoder_",
+    "dataset_transformer_",
 }
 
 keras_regressor_base_meta_set = {
@@ -36,6 +37,7 @@ keras_regressor_base_meta_set = {
     "target_type_",
     "feature_encoder_",
     "target_encoder_",
+    "dataset_transformer_",
 }
 
 


adriangb:
@stsievert if you are able to, a review of this new interface would be much appreciated. The best place to start is probably [section 6 and 7 of this notebook](https://www.adriangb.com/scikeras/refs/pull/167/merge/notebooks/DataTransformers.html#6.-Ragged-datasets-with-tf.data.Dataset).

scikeras/utils/transformers.py
```diff
@@ -390,3 +407,31 @@ def get_metadata(self):
...
+
+    def transform(
+        self, data: Tuple[np.ndarray, Optional[np.ndarray], Optional[np.ndarray]]
+    ) -> Tuple[np.ndarray, Union[np.ndarray, None], Union[np.ndarray, None]]:
+        X, y, sample_weight = data
+        if self.class_weight is None or y is None:
+            return (X, y, sample_weight)
+        sample_weight = 1 if sample_weight is None else sample_weight
+        sample_weight *= compute_sample_weight(class_weight=self.class_weight, y=y)
+        return (X, y, sample_weight)
```
adriangb, line 437: There is another option here, that may be cleaner, which would be to add another dependency injection point that runs in "parallel" to `feature_encoder` and `target_encoder` specifically for `sample_weight`. But since it also makes sense that one would need at least `y` to process `sample_weight`, it might be a bit redundant.

docs/source/notebooks/DataTransformers.md
```diff
@@ -43,6 +43,8 @@ In this notebook, we explore how to reconcile this functionality with the sklear
 * [5. Multidimensional inputs with MNIST dataset](#5.-Multidimensional-inputs-with-MNIST-dataset)
```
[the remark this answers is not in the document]
stsievert: I don't know if I'd include the psuedo-code in this notebook: that seems like it should be in the API documentation, not an example.

docs/source/advanced.rst
```diff
@@ -178,11 +178,50 @@ This is basically the same as calling :py:func:`~scikeras.wrappers.BaseWrapper.g
...
+limited options, SciKeras introduces "data transformers". These are really just
+dependency injection points where you can declare custom data transformations,
+for example to split an array into a list of arrays, join `X` & `y` into a `Dataset`, etc.
+In order to keep these transformations in a familiar format, they are implemented as
+sklearn-style transformers. You can think of this setup as an sklearn Pipeline:
+
+.. code-block::
+
+                                             ↗ feature_encoder ↘
+    your data → sklearn-ecosystem → SciKeras                    dataset_transformer → Keras
```
stsievert, line 196: I think this diagram is really useful.

Would this be a better diagram?


```
                                   ↗ feature_encoder ↘
    SciKeras.fit(features, labels)                    dataset_transformer → Keras.fit(dataset)
                                   ↘ target_encoder  ↗ 
```

docs/source/advanced.rst
```diff
@@ -178,11 +178,50 @@ This is basically the same as calling :py:func:`~scikeras.wrappers.BaseWrapper.g
...
+As you can see, there are 2 stages of data transformations within SciKeras:
+
+- Target/Feature transformations:
+    - feature_encoder: Handles transformations to the features (`X`). This can be used 
+      to implement multi-input models.
+    - target_encoder: Handles transformations to the target (`y`). This can be used
+      to implement non-int labels (eg: strings) as well as mutli-output models.
+- Whole dataset transformations:
+    - dataset_transformer: This is the last step before passing the data to Keras.
+      It can be used to implement conversion to a `Dataset`, amongst other things.
```
stsievert, line 209:
```suggestion
      By default, this transformation does nothing and passes the features/labels aka `X`/`y` to [Keras.fit].
      It's signature if `dataset_transformer.fit_transform(X, y)`
      It can be used for any input that [Keras.fit] supports, including to implement conversion to a `Dataset`, or any datatype that [Keras.fit] supports.
      
[Keras.fit]:https://keras.io/api/models/model_training_apis/#fit-method
```

docs/source/advanced.rst
```diff
@@ -178,11 +178,50 @@ This is basically the same as calling :py:func:`~scikeras.wrappers.BaseWrapper.g
...
+
+`feature_encoder` and `target_encoder` are run before building the Keras Model,
```
stsievert, line 211: Style nit:

```suggestion
``feature_encoder`` and ``target_encoder`` are run before building the Keras Model,
```

This is restructured text, not markdown. This is also the case elsewhere in this document.

docs/source/notebooks/DataTransformers.md
```diff
@@ -543,3 +549,297 @@ clf.fit(x_train, y_train)
...
 ```
+
+## 6. Ragged datasets with tf.data.Dataset
+
+SciKeras provides a third dependency injection point that operates on the entire dataset: X, y & sample_weight. This `dataset_transformer` is applied after `target_transformer` and `feature_transformer`. One use case for this dependency injection point is to transform data from tabular/array-like to the `tf.data.Dataset` format, which only requires iteration. We can use this to create a `tf.data.Dataset` of ragged tensors.
+
+Note that `dataset_transformer` should accept a single **3 element tuple** as its argument and return value:
+
+```python
+help(KerasClassifier.dataset_transformer)
```
stsievert, line 560: Style nit: link to the docs on this.

docs/source/notebooks/DataTransformers.md
```diff
@@ -543,3 +549,297 @@ clf.fit(x_train, y_train)
...
+The use of a 3 element tuple allows you to chain transformers with this same interface using a Scikit-Learn Pipeline, as you will see below.
+
+When you return a tuple like `(tf.data.Dataset(...), None, None)`, SciKeras will pass the data untouched to `Model.fit` like `Model.fit(x=tf.data.Dataset(...), y=None, sample_weight=None)`. You can process these arguments in any way you like, as long as Keras accepts them, SciKeras will not complain.
+
+Let's start by defining our data. We'll have an extra "feature" that marks the observation index, but we'll remove it when we deconstruct our data in the transformer.
+
+```python
+feature_1 = np.random.uniform(size=(10, ))
+feature_2 = np.random.uniform(size=(10, ))
+obs = [0] * 3 + [1] * 2 + [2] * 1 + [3] * 2 + [4] * 2
```
stsievert, line 572:
```suggestion
obs = [0, 0, 0, 1, 1, 2, 3, 3, 4, 4]
```

docs/source/notebooks/DataTransformers.md
```diff
@@ -543,3 +549,297 @@ clf.fit(x_train, y_train)
...
+        if y is not None:
+            # y should be a list of arrays, as split up by MultiOutputTransformer
+            sample_weights = dict()
+            for output_num, (output_name, output_data) in enumerate(zip(self.output_names, y)):
+                # class_weight is expected to be indexable by output_number
+                # see https://scikit-learn.org/stable/modules/generated/sklearn.utils.class_weight.compute_sample_weight.html
+                # Note that it is trivial to change the expected format to match Keras' ({output_name: weights, ...})
+                # see https://github.com/keras-team/keras/issues/4735#issuecomment-267473722
+                cls_wt_out = class_weight[output_num]
+                sample_weights[output_name] = compute_sample_weight(cls_wt_out, output_data)
```
stsievert, line 753: Suggestion:

``` python
sample_weights = {
    compute_sample_weight(class_weight[output_num], output_data)
    for output_num, (output_name, output_data) in enumerate(zip(self.output_names, y))
}
```

docs/source/advanced.rst
adriangb: I do like the addition of `.fit`.  My only worry is that users might think that SciKeras always creates a `tf.data.Dataset`, which is not the case, by default it gives numpy arrays to `Keras.fit`. Do you think `Keras.fit(dataset or np.array)` makes that clear? It could also be dicts of lists or something, but that's at least more uncommon.

docs/source/advanced.rst
adriangb: Thanks for pointing it out. Will fix.

docs/source/notebooks/DataTransformers.md
adriangb: Instead of calling `help(...)` or in addition to?

docs/source/notebooks/DataTransformers.md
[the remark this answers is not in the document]
adriangb: The goal was to introduce the general structure/plan of what the notebook was going to introduce without getting bogged down in the details by introducing a fully working version.

I initially put the `if False:`s in there to avoid errors when executing, which was probably just lazy. I can probably remove all of the `if False:` by just making sure all of the pseudocode is valid python, or just marking the cell as non-executable (which is possible). Would this resolve the problem? I am not sure I am understanding how mospaced font plays into this.

docs/source/notebooks/DataTransformers.md
```diff
@@ -43,6 +43,8 @@ In this notebook, we explore how to reconcile this functionality with the sklear
 * [5. Multidimensional inputs with MNIST dataset](#5.-Multidimensional-inputs-with-MNIST-dataset)
   * [5.1 Define Keras Model](#5.1-Define-Keras-Model)
   * [5.2 Test](#5.2-Test)
+* [6. Ragged datasets with tf.data.Dataset](#6.-Ragged-datasets-with-tf.data.Dataset)
+* [7. Multi-output class_weight](#7.-Multi-output-class_weight)
 
 ## 1. Setup
```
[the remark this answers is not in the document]
adriangb: I agree. I will coalesce it into `advanced.rst`.

docs/source/advanced.rst
stsievert: What can `dataset_transformer` return? Only datasets/ndarrays? Or does it support the other arguments of Keras.fit?

adriangb: Anything that `Keras.fit` will accept. Internally, it looks something like this:

```python3
X, y, sample_weight = dataset_transformer.fit_transform((X, y, sample_weight))
model.fit(x=X, y=y, sample_weight=sample_weight)  # aka Keras.fit
```

docs/source/notebooks/DataTransformers.md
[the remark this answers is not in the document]
stsievert: > introduce the general structure/plan of what the notebook was going to introduce without getting bogged down in the details by introducing a fully working version.

I think that goal can be accomplished without any psuedo-code. "This notebook will detail how to do the following: [list]."

Marking the cell as non-executable would resolve the complaint about `if False`

docs/source/advanced.rst
adriangb: Maybe `Keras.fit(data)` is a good way to specify this? That way there's not confusion in interpreting `dataset` as `tf.dataDataset`. I can also add a small code block like in https://github.com/adriangb/scikeras/pull/167#discussion_r563994781 if that helps explain it.

docs/source/notebooks/DataTransformers.md
stsievert: I'd definitely link to the docs. 

I'm not sure calling `help` does anything meaningful. It outputs the docstring, right? I'm not sure that's super meaningful, unless you refer to a specific point later in this notebook.

adriangb: I think just linking to the docstring within the docs (like [this](https://www.adriangb.com/scikeras/refs/heads/master/generated/scikeras.wrappers.KerasClassifier.html#scikeras.wrappers.KerasClassifier.feature_encoder)) should suffice. The information on the signature and such is important to understand the notebook, but I wanted to avoid duplicating the text. I'll rewrite it as a 1-2 sentence explanation in the notebook + a link to the docs.

docs/source/notebooks/DataTransformers.md
```diff
@@ -43,6 +43,8 @@ In this notebook, we explore how to reconcile this functionality with the sklear
 * [5. Multidimensional inputs with MNIST dataset](#5.-Multidimensional-inputs-with-MNIST-dataset)
   * [5.1 Define Keras Model](#5.1-Define-Keras-Model)
   * [5.2 Test](#5.2-Test)
+* [6. Ragged datasets with tf.data.Dataset](#6.-Ragged-datasets-with-tf.data.Dataset)
+* [7. Multi-output class_weight](#7.-Multi-output-class_weight)
```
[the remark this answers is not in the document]
adriangb: I added a similar note to the top of the notebook. Thank you for the suggestion.

docs/source/notebooks/DataTransformers.md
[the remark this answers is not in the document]
adriangb: I deleted most of the pseudocode and moved the rest to `advanced.rst`.

That said, for future reference, marking one of these markdown code cells as non executable is as easy as:

```
    ```python .noeval
    # code here
    ```
```

docs/source/notebooks/DataTransformers.md
adriangb: Incorporated

docs/source/notebooks/DataTransformers.md
adriangb: Incorporated, thank you

docs/source/advanced.rst
adriangb: I _think_ I got them all.

docs/source/advanced.rst
adriangb: I ended up re-organizing this entire section, I think I incorporated this part [here](https://www.adriangb.com/scikeras/refs/pull/167/merge/advanced.html#whole-dataset-manipulation-via-data-transformer), but let me know if it is not clear.

adriangb:
@stsievert thank you for the review, the feedback was very valuable, I think I was able to incorporate most of the smaller suggestions.

For the larger stuff, I moved most of the pseudocode and general background out of the notebook and into the main docs, and added some links back to the docs instead of duplicate information, as you suggested.

Having read the documentation, do you think this `data_transformer` interface makes sense? And in particular, do you feel that using it to handle `class_weight` is problematic?

docs/source/advanced.rst
```diff
@@ -178,26 +178,126 @@ This is basically the same as calling :py:func:`~scikeras.wrappers.BaseWrapper.g
...
+.. code-block::
+
+                                   ↗ feature_encoder ↘
+    SciKeras.fit(features, labels)                    dataset_transformer → keras.Model.fit(data)
+                                   ↘ target_encoder  ↗ 
+
+
+Within SciKeras, this is roughly implemented as follows:
+
+.. code:: python
```
stsievert, line 202: I think a lot of this code block is unnecessary. I'd use this:

``` python
class PseudoBaseWrapper:
    def fit(self, X, y, sample_weight):
        self.target_encoder_ = self.target_encoder.fit(y)
        self.feature_encoder_ = self.feature_encoder.fit(X)
        self.model_ = self._build_keras_model()
        X, y, sample_weight = self.dataset_encoder.fit(
            self.feature_encoder_.transform(X), self.target_encoder_.transform(y), sample_weight
        )
        self.model_.fit(X=X, y=y, sample_weight=sample_weight)  # tf.keras.Model.fit
        return self

    def predict(self, X):
        X, _, _ = self.dataset_transformer_.fit_transform((X, None, None))
        y_pred = self.model_.predict(X)
        return self.target_encoder_.inverse_transform(y_pred)
```

docs/source/advanced.rst
```diff
@@ -178,26 +178,126 @@ This is basically the same as calling :py:func:`~scikeras.wrappers.BaseWrapper.g
...
+            return RegressorTargetEncoder()
+
+
+One important thing to note is that ``feature_encoder`` and ``target_encoder``
+are run before building the Keras Model, while ``data_transformer`` is run after
+the Model is built. This means that the former two will not have access to the Model
+(eg. to get the number of outputs) but *will* be able to inject data into the model building
+function (more on this below). On the other hand,
+``data_transformer`` *will* get access to the built Model, but it cannot pass any data to model building
+function.
```
stsievert, line 273: > One important thing to note is that feature_encoder and target_encoder are run before building the Keras Model, while data_transformer is run after the Model is built.

This isn't clear in the example (though it's clear in the edit I've made).

> This means that the former two will not have access to the Model
(eg. to get the number of outputs) but *will* be able to inject data into the model building
function (more on this below). On the other hand,
``data_transformer`` *will* get access to the built Model, but it cannot pass any data to model building
function.

I would only say "the output of `dataset_transformer` get directly passed to tf.keras.Model.fit through self.model_.fit." All the about the number of outputs/etc is a confusing. I think I'd refer directly to the advanced examples.

docs/source/advanced.rst
```diff
@@ -229,11 +329,61 @@ type, and implements basic handling of the following cases out of the box:
...
+The default implementations are available for use as :py:class:`scikeras.utils.transformers.ClassifierLabelEncoder`
+and :py:class:`scikeras.utils.transformers.RegressorTargetEncoder` for
+:py:class:`scikeras.wrappers.KerasClassifier` and :py:class:`scikeras.wrappers.KerasRegressor` respectively.
 
-For a complete examples implementing custom data processing, see the examples in the
-:ref:`tutorials` section.
+As per the table above, if you find that your target is classified as
+``"multiclass-multioutput"`` or ``"unknown"``, you will have to implement your own data processing routine.
+
+Whole dataset manipulation via data_transformer
```
stsievert, line 340: I think this section should go directly below the paragraph on dataset transformers.

docs/source/advanced.rst
```diff
@@ -229,11 +329,61 @@ type, and implements basic handling of the following cases out of the box:
...
++++++++++++++++++++++++++++++++++++++++++++++++
+
+This is the last step before passing the data to Keras, and it allows for the greatest
+degree of customization because SciKeras does not make any assumptions about the output data
+and passes it directly to :py:func:`tensorflow.keras.Model.fit`.
+Its signature is ``dataset_transformer.fit_transform((X, y, sample_weight))``,
```
stsievert, line 346: It's signature is this:

``` python
from sklearn.base import BaseEstimator

class DatasetTransformer(BaseEstimator):
    def fit(self, X, y, sample_weight=None) -> "DatasetTransformer":
        ...
        return self

    def transform(self, X, y, sample_weight):  # return a valid input for keras.Model.fit
        ...
        return X, y, sample_weight  # option 1
        return tensorflow_dataset  # option 2
```

docs/source/advanced.rst
stsievert: I'm also unsure about the level of nesting here. It looks like "multi-input" and "whole dataset" are on the same level, something I wouldn't have expected with the titles.

It might be worth collapsing this section into the "data transformers" section. This section really only mentions the signature.

docs/source/advanced.rst
adriangb: I may have design it wrong, and I'm not a fan of the tuple input either, but I believe it is required so that these can be chained in a pipeline.

With the signature you are proposing, this fails:

```python3
from sklearn.base import BaseEstimator
from sklearn.pipeline import make_pipeline

class DatasetTransformer(BaseEstimator):
    def fit(self, X, y, sample_weight=None) -> "DatasetTransformer":
        return self

    def transform(self, X, y, sample_weight):  # return a valid input for keras.Model.fit
        return X, y, sample_weight

p = make_pipeline(DatasetTransformer(), DatasetTransformer())

X = [1]
y = [1]

p.fit_transform(X, y)
```

If instead you accept a tuple as your input, then you can chain them:

```python3
class DatasetTransformer(BaseEstimator):
    def fit(self, X, y, sample_weight=None) -> "DatasetTransformer":
        return self

    def transform(self, data):  # return a valid input for keras.Model.fit
        X, y, sample_weight = data
        return X, y, sample_weight

p = make_pipeline(DatasetTransformer(), DatasetTransformer())

X = [1]
y = [1]

p.fit_transform((X, y, None))
```

Am I missing something?

docs/source/advanced.rst
adriangb: I agree that we can remove the `@property` stuff. I'll implement that.

docs/source/advanced.rst
adriangb: > This isn't clear in the example (though it's clear in the edit I've made).

Thank you for pointing that out! The example is much better with this detail.

> I would only say

That seems reasonable, I'll cut it down to ~ what you are suggesting.

docs/source/advanced.rst
adriangb: I'll move it as you suggest. It then becomes duplicated with the paragraph discussed in https://github.com/adriangb/scikeras/pull/167#discussion_r565614539, which warrants removing the latter.

docs/source/advanced.rst
stsievert: Thanks. Nothing I say hard and fast, all my comments are comments are illustrations. If a change to be hard and fast, I'll submit a PR.

docs/source/advanced.rst
stsievert: With the example, I only meant "illustrate the signature with code, not with words." I am unsure what's happening with tuple inputs.

adriangb: Oh I see. Yes, that is a good suggestion.

stsievert: Maybe it'd be clearer to have the user return a dict. I'd also modify the tuple input to this transform; why not pass keyword arguments?

``` python
ValidKerasInput = Union[np.ndarray, tf.Dataset, ...]

def transform(
    self,
    X: ValidKerasInput,
    y: ValidKerasInput,
    sample_weight: Optional[ValidKerasInput]=None
) -> dict[str, ValidKerasInput]:
    ...
    return {"X": X, "y": y, "sample_weight": None}
```

adriangb: That signature would work (and is ideal) for a single transformer. I think the issue is that sklearn assumes the signature `transform(X)` (i.e. a single argument), so to be compatible with the input must be a single argument. And to chain, the input and output must be identical.

That said, we could swap from a tuple to a dict. As long as it passes this test it should work:

```python3
p = make_pipeline(DatasetTransformer(), DatasetTransformer())
p.fit_transform((X, y, sample_weight)) # or {"X": X, "y"....}
```

stsievert: > I think the issue is that sklearn assumes the signature transform(X) (i.e. a single argument), so to be compatible with the input must be a single argument. And to chain, the input and output must be identical.

Yes, that's the formal Scikit-learn API specification. But, that specification is not enforced. I wouldn't complain if you broke that specification: I can't imagine users will chain `DatasetTransformer`s together, and I think multiple inputs will enhance usability.

Also, conforming to API specifications is a good practice, so I won't complain if you follow the API specification.

adriangb: It's not just about the API specification, in my very limited testing I already found use cases for chaining, like [this part of the notebooks](https://www.adriangb.com/scikeras/refs/pull/167/merge/notebooks/DataTransformers.html#5.-Ragged-datasets-with-tf.data.Dataset):


```python3
class RaggedClassifier(KerasClassifier):

    @property
    def dataset_transformer(self):
        t1 = FunctionTransformer(ragged_transformer)
        t2 = super().dataset_transformer  # ClassWeightDataTransformer
        t3 = FunctionTransformer(dataset_transformer)
        return make_pipeline(t1, t2, t3)
```

adriangb: I was able to remove `dummy`/the second parameter by adding a `"passthrough"` step in the pipeline. This is still a quirk, but a lesser evil (I think).

stsievert: I'm not sure I understand. It took a while to understand the `passthrough` parameter (I think I understand it). I can understand a tuple/dict right away, but the "passthrough" option requires some thought and digging.

I'm a fan of passing a dict around:

``` python
p = make_pipeline(DatasetTransformer(), DatasetTransformer())
data = {"X": X, "y": y, "sample_weight": sample_weight}
p.fit(data)
p.transform(data)
p.fit_transform(data)
```

I slightly prefer passing a dict around because the keys aren't fixed, and a tuple has a static length.

adriangb: A dict sounds good, thank you for the suggestion. As a side note, I dug into the issue a bit more, it seems that being able to modify only X (vs. y and sample_weight like we do here) is a [known design flaw](https://github.com/scikit-learn/scikit-learn/issues/3855#issuecomment-63213164) in sklearn. This makes me feel a bit more comfortable "adapting" the API in this fashion.

Regarding the `"passthrough"`/`dummy` thing: from the [Pipeline docs](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html), the final estimator _must_ implement `fit`. And `fit` is expected to have the signature `fit(X, y)` and is called like that ([here](https://github.com/scikit-learn/scikit-learn/blob/b3ea3ed6a09fe774dfc5160a65172b1bacbb2a82/sklearn/pipeline.py#L346)). But since we are including `y` in `data`, it does not make sense to have the signature `fit(data, y=None)` and then override `y` in the body of the function. Hence why I originally renamed the parameter from `y` to `dummy` to avoid confusion. The alternative, which is a fix specifically for pipelines but may leave room for issues elsewhere, is to use the better signature of `fit(data)` but force the last estimator in the pipeline to be `"passthrough"`/`None`. This means that all of these transformers are "known" to be transformers and so they are called with just `X` (here).

I'm a bit torn here, on the one hand, I don't like having a parameter called `dummy`. On the other hand, I think the `"passthrough"` solution is more fragile (and it isn't very nice either).

stsievert: > ` fit(data, y=None)`

Why not `fit(data, **ignored)` or `fit(data, **empty)`? If you want, you can even write `assert len(empty) == 0`.

adriangb: I agree. Let me try implementing this tomorrow.

For the dict, do you envision passing/returning `None` or just not including the key?

Also, a minor nit, but sklearn uses `X` (capital) while Keras uses `x` (lower). There's something to be said for using `x`, since it allows for `self.model_.fit(**data, ...)`. Do you have a preference for `X` or `x`?

stsievert: I think it's easiest to say "the keys in the returned dictionary should be valid keywords to keras.Model.fit".

Let's take a transformer that does nothing and returns the input, in-modified.

Then, `x` should be preferred because it's a valid keyword argument to keras.Model.fit. I suspect that answers the no key question.

adriangb: If we are going to go with dicts, what are your thoughts on passing all `Model.fit`/`Model.predict` `**kwargs`?

The pro is that it would allow nearly arbitrary user logic injection. For example, #131 could be resolved by creating a transformer that siphons off validation data from `x` & `y` into `validation_dataset`. This would be done after applying target encoding, etc. so it should be compatible with all of sklearn and Keras.

The con is that I think the interface could be messy.

stsievert: How would it be messy? Arbitrary user logic sounds like a good feature for something basic like this.

> If we are going to go with dicts, what are your thoughts on passing all Model.fit/Model.predict **kwargs?

Doesn't a user-passed kwarg already passed to `fit`? I'm thinking of `scikeras_model.fit(...,  batch_size=42)`. That's in `kwargs` and gets passed to `model_.fit`, right?

adriangb: > How would it be messy?

It's minor, but it would now be a dict with dozens of keys, some values being arrays, etc. It can easily be documented, but I can see how at first glance it might be more confusing than a 3 element tuple or a 3 key dict.

> Doesn't a user-passed kwarg already passed to fit?

Yes, but these parameters can't be calculated based on the data, which becomes especially problematic if doing cross-valdation and such. This is the basic problem in #131.

Of course, if users want to pass a fixed parameter like `batch_size=42` they can do so via `fit` (or better yet, via the constructor).

scikeras/wrappers.py
```diff
@@ -755,20 +797,24 @@ def _initialize(
...
-        feature_meta = getattr(self.feature_encoder, "get_metadata", dict)()
+        feature_meta = getattr(self.feature_encoder_, "get_metadata", dict)()
         vars(self).update(**feature_meta)
 
         self.model_ = self._build_keras_model()
 
-        return X, y
+        self.dataset_transformer_ = self.dataset_transformer.fit((X, y, sample_weight))
+        dataset_meta = getattr(self.dataset_transformer_, "get_metadata", dict)()
+        vars(self).update(**dataset_meta)
```
stsievert, line 813: :-1:

docs/source/advanced.rst
stsievert: > It's minor, but it would now be a dict with dozens of keys,

Why would it have to be dozens of keys? I'm thinking of this code:

``` python
# line 885 of wrappers.py
transform_kwargs = self.dataset_transformer_.transform(x=X, y=y, sample_weight=sample_weight)
kwargs.update(transform_kwargs)
self._fit_keras_model(**kwargs)
```

scikeras/wrappers.py
stsievert:
```suggestion
        self.dataset_meta_ = dataset_meta
```

adriangb: Which part of these 3 lines are you 👎 on? This code hasn't stayed up to date with our discussions, so I'm not sure if it's the tuple part or something else.

adriangb: Ah, sorry, I see now we cross posted.

stsievert: I think this comment is tangential to the other discussion. I would rather consolidate similar keys into one dictionary. I'm not a fan of blindly assigning attributes to objects.

adriangb: Agreed. We actually weren't using this at the moment, and there's no reason it can't be added at a later date if a good feature requires it, so I was able to remove it for now. Does that sound like it resolves this problem?

docs/source/advanced.rst
adriangb: I think that implementation in https://github.com/adriangb/scikeras/pull/167#discussion_r576512695 is not compatible with the sklearn ecosystem, including pipelines, which I think are in important feature. This would be fine:

```python3
self.dataset_transformer_.transform(dict(x=X, y=y, sample_weight=sample_weight))
```

Regarding the multiple keys, what I was thinking was to pass _all_ `Model.fit` keys. In `_fit_keras_model`:

https://github.com/adriangb/scikeras/blob/6eee3c4a19b28694758e02aab27af9222ff82db8/scikeras/wrappers.py#L480-L494

This means that users can not only add keys, but also modify existing keys, to their liking.

Regarding the `dozens of keys` comment, I wasn't referring to the keys assigned via `kwargs.update(transform_kwargs)` above, I was more thinking that when a user creates a transformer like:

```python
def transform(data: Dict[str, Any) -> Dict[str, Any]:
    len(data.keys())  # many keys, array-values, etc.
    ...

# for context:
class SomeCustomEstimator(BaseWrapper):
    @property
    def data_transformer(self):
        return FunctionTransformer(transformer)
```

scikeras/wrappers.py
stsievert: Yeah, I think so. I'm not opposed to setting the metadata about the dataset in a single attribute:

``` python
self.dataset_meta_ = getattr(self.dataset_transformer_, "get_metadata", dict)()
```

docs/source/advanced.rst
stsievert: > This means that users can not only add keys, but also modify existing keys, to their liking.

Yes, I think that's a good idea. It also leads to simple documentation (something like "the dataset transformer is passed a dict. Normally, this dictionary is unmodified and passed directly to keras.Model.fit. If desired, it's possible to modify this dictionary through a Scikit-learn transformation: [example]."

adriangb: Great. I'm going to loop back to #131 and #159 to see if the issue reporters have any input on this implementation.

scikeras/wrappers.py
adriangb: Since I removed calling `get_metadata` for `dataset_transformer_`, I'm going to resolve this comment.

PR #176, opened 2021-01-23
Typo fix in DataTransformers.ipynb

data-hound:
Under Section 2, Multiple Inputs
 
```
class MultiOutputTransformer(BaseEstimator, TransformerMixin):
    ...

class MultiOutputClassifier(KerasClassifier):

    @property
    def feature_encoder(self):
        return MultiInputTransformer(...)
```

  changed to
  
```
class MultiInputTransformer(BaseEstimator, TransformerMixin):
    ...


class MultiInputClassifier(KerasClassifier):

    @property
    def feature_encoder(self):
        return MultiInputTransformer(...)
```


adriangb:
Thank you for the bug report! This has been noted and will be fixed.

.pre-commit-config.yaml
```diff
@@ -1,12 +1,12 @@
-repos:
```
adriangb, line 1: Are the changes here intentional?

data-hound: Nope. After I ran the pre-commit test, and everything was fine, I was asked to stage this file for commit as well, and so I did.

adriangb: Hmm, not sure what happened there. Would you mind reverting this change then?

adriangb:
I'm actually in the middle of a pretty large overhaul of the docs in #174, the main goal being to make it easier to edit notebooks. Would you mind waiting until I merge #174 and then we can use this change as a case study for the effectiveness of the new system?

data-hound:
I see the #174 has been merged. I will make the changes then.

adriangb:
Awesome, thank you! Please let me know if you encounter any issues with the new setup.

adriangb:
Hey @data-hound it looks like the way I set up CI won't work for forks 🤦 . I think I'll be able to get it fixed, but it might take a couple of days. That is to say, don't worry about the CI failure, and sorry for wasting your time with broken CI.

adriangb:
I was able to make the changes. Thank you very much @data-hound !

data-hound:
> I was able to make the changes. Thank you very much @data-hound !

Mention Not! 
Seems you did the rectification for the pre-commit-config.yaml in the branch as well - Thanks for that!

PR #180, opened 2021-01-26
MAINT: remove unecessary sample_weight logic

adriangb:
I believe the test that was failing here was resolved in https://github.com/adriangb/scikeras/pull/152.

In hindsight, I think the original issue was that 1 epoch of training probably wasn't enough to see the effect of sample_weight in the way the test is looking for.


adriangb:
I've run this >4 times locally and 2x in CI and can't seem to get it to fail, so it's at least not _very_ flakey now. Hence I'm going to merge this and assume the diagnosis is correct.

PR #186, opened 2021-02-01
TST: remove precision hack in tests now that we use float64 in tests

adriangb:


stsievert:
SciKeras still uses float32 by default, right?

adriangb:
Yes, there are just a handful of sklearn checks that require the higher precision

stsievert:
Good. float32 has more FLOPS than float64 because it requires less memory. Plus, float32 is plenty accurate enough – even [NASA only 15 digits for pi][1] (aka a [machine epsilon][2] of 1e-15) for Voyager, which is currently 1.25 billion miles from earth.

[2]:https://en.wikipedia.org/wiki/Machine_epsilon
[1]:https://www.jpl.nasa.gov/edu/news/2016/3/16/how-many-decimals-of-pi-do-we-really-need/

adriangb:
Great info.

Really, these tests shouldn't be this restrictive, and I've pinged the sklearn team about it, but for now, this is the best we can do

adriangb:
Having run this several times, it seems to be stable, so I will go ahead and merge. That said, as these tests have been an ongoing flaky issue, it is hard to tell if this is a true fix, but I will keep an eye on them.

PR #194, opened 2021-02-15
Implement batch_size=-1

adriangb:


stsievert:
I'm still a fan of keeping `**kwargs` in `model.fit` as expressed in https://github.com/adriangb/scikeras/issues/159#issuecomment-753750516. This PR is tangential to that.

Style nit: Skorch `batch_size=-1` for the same functionality this PR implements ([Skorch docstring][1]). I think that's a little better because it mirrors the negative list indexing.

[1]:https://github.com/skorch-dev/skorch/blob/aa99dbc3af6925f8e77679a0366adff5d75c9b66/skorch/net.py#L115-L117

adriangb:
> I'm still a fan of keeping `**kwargs` in `model.fit` as expressed in [#159 (comment)](https://github.com/adriangb/scikeras/issues/159#issuecomment-753750516). This PR is tangential to that.

Agreed, I don't plan on removing `**kwargs` just because this is implemented (if it is). Also tangential, but I'm going to punt a final decision on `**kwargs` down the road since I think some more use cases / user input would be good before making a decision either way.

`-1` should work as well, I do like that it avoids introducing a new type to the parameter (str). Thanks for pointing this out.

adriangb:
I think this change is pretty non-controversial, especially since Skorch already implements the exact same feature.

I'm going to leave this open a couple more days for feedback and then if there are no objections I'll go ahead and merge it.

PR #197, opened 2021-02-15
DOC/TST: Use r2 as a metric

### LINKED ISSUE #196: DOC/TST: Clarify usage of r_squred
`KerasRegressor.r_squared` is supposed to be used as a metric with Keras, but it is not tested as such and it is incorrectly used as a loss function in some places. This is very bad. 2 things are required to fix this:
1. Add a test that checks that it works as a metric.
2. Replace every usage of it as a loss function with mean squared error, including in notebooks.

adriangb:
Fixes #196

### DIFF
--- a/docs/source/notebooks/Basic_Usage.md
+++ b/docs/source/notebooks/Basic_Usage.md
@@ -182,7 +182,7 @@ def get_reg(meta, hidden_layer_sizes, dropout):
 
 ### 3.3 Defining and training the neural net regressor
 
-Training a regressor is almost the same as training a classifier. Mainly, we use `KerasRegressor` instead of `KerasClassifier` (this is the same terminology as in `sklearn`). We also change the loss function to `KerasRegressor.r_squared`. SciKeras provides this loss function because most of the `sklearn` ecosystem expects `R^2` as the loss function, but Keras does not have a default implementation.
+Training a regressor is almost the same as training a classifier. We use `KerasRegressor` instead of `KerasClassifier` and add `KerasRegressor.r_squared` as a metric. SciKeras provides this metric function because most of the `sklearn` ecosystem scores regressors with `R^2`, but Keras does not have a default implementation.
 
 ```python
 from scikeras.wrappers import KerasRegressor
@@ -190,7 +190,8 @@ from scikeras.wrappers import KerasRegressor
 
 reg = KerasRegressor(
     model=get_reg,
-    loss=KerasRegressor.r_squared,
+    loss="mse",
+    metrics=[KerasRegressor.r_squared],
     hidden_layer_sizes=(100,),
     dropout=0.5,
 )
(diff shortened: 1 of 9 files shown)


scikeras/wrappers.py
```diff
@@ -1611,7 +1611,7 @@ def target_encoder(self):
     @register_keras_serializable()
     def r_squared(y_true, y_pred):
         """A simple Keras implementation of R^2 that can be used as a Keras
-        loss function.
+        metric function.
```
stsievert, line 1614: Are lower values better for this metric?

docs/source/notebooks/Basic_Usage.md
```diff
@@ -182,15 +182,16 @@ def get_reg(meta, hidden_layer_sizes, dropout):
 
 ### 3.3 Defining and training the neural net regressor
 
-Training a regressor is almost the same as training a classifier. Mainly, we use `KerasRegressor` instead of `KerasClassifier` (this is the same terminology as in `sklearn`). We also change the loss function to `KerasRegressor.r_squared`. SciKeras provides this loss function because most of the `sklearn` ecosystem expects `R^2` as the loss function, but Keras does not have a default implementation.
+Training a regressor is almost the same as training a classifier. We use `KerasRegressor` instead of `KerasClassifier` and add `KerasRegressor.r_squared` as a metric. SciKeras provides this metric function because most of the `sklearn` ecosystem scores regressors with `R^2`, but Keras does not have a default implementation.
```
stsievert, line 185:
```suggestion
Training a regressor has nearly the same data flow as training a classifier. The differences include using `KerasRegressor` instead of `KerasClassifier` and adding `KerasRegressor.r_squared` as a metric. Most of the Scikit-learn regressors use the coefficient of determination or R^2 as a metric function, which measures correlation between the true labels and predicted labels.
```

tests/mlp_models.py
```diff
@@ -66,7 +66,7 @@ def dynamic_regressor(
     n_features_in_ = meta["n_features_in_"]
     n_outputs_ = meta["n_outputs_"]
 
-    compile_kwargs["loss"] = compile_kwargs["loss"] or KerasRegressor.r_squared
+    compile_kwargs["loss"] = compile_kwargs["loss"] or "mse"
```
stsievert, line 69: Yikes. This fix is needed because minimizing R^2 is the same as maximizing MSE.

scikeras/wrappers.py
adriangb: Good point. How about adding:
> Larger values indicate a better fit, with 1.0 representing a perfect fit.

tests/mlp_models.py
adriangb: Yep! A huge upsie. It wasn't caught until now because (1) this is in tests, so no users were experiencing this and (2) none of the tests were actually checking scores.

PR #198, opened 2021-02-16
MAINT: remove deprecation warning for **kwargs

adriangb:
This PR removes the deprecation warning for **kwargs, hence making it a feature that will stay.

### DIFF
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -83,7 +83,8 @@ If you want to change the parameters on a live instance, you can do:
    clf.fit(...)
 
 Functionally, this is the same as passing these parameters to ``fit``, just with one more function call.
-This is what Scikti-Learn does in the background for hyperparameter tuning.
+This is much more compatible with the Scikit-Learn API.
+In fact, this is what Scikti-Learn does in the background for hyperparameter tuning.
 
 Renaming of ``build_fn`` to ``model``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -49,15 +49,12 @@ pass your loss function to the constructor:
 Variable keyword arguments in fit and predict
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Keras supports a large number of variable keyword arguments keyword arguments (commonly referred to as
-``**kwargs``) for ``fit`` and ``predict``. SciKeras supports these as well,
-but passing arguments in this way means that they will not be hyperparameter-tunable, can't be
-pickled, etc.
+Keras supports a variable keyword arguments (commonly referred to as ``**kwargs``) for ``fit`` and ``predict``.
+Scikit-Learn on the other hand does not support these arguments, and using them is largely incompatible with the Scikit-Learn ecosystem.
+As a compromise, SciKeras supports these arguments, but we recommended that you set parameters using the constructor
+or ``set_params`` for first-class SciKeras support.
 
-In general, it is better to declare these parameters in the constructor of your model.
-While this is not possibel with all parameters, it is not a problem for the vast majority of them.
-
-As an example, we declare ``batch_size`` in the constructor:
+For example, to declare ``batch_size`` in the constructor:
 
 .. code:: diff
 
@@ -70,7 +67,18 @@ Or to declare separate values for ``fit`` and ``predict``:
 
 .. code:: python
 
-   clf = KerasClassifier(fit__batch_size=32, predict__batch_size=10000)
+   clf = KerasClassifier(..., fit__batch_size=32, predict__batch_size=10000)
+
+If you want to change the parameters on a live instance, you can do:
+
+.. code:: python
+
+   clf = KerasClassifier(...)
+   clf.set_params(fit__batch_size=32, predict__batch_size=10000)
+   clf.fit(...)
+
+This effectively gives you the same result as passing ``**kwargs``, just with
+one more function call.
 
 Renaming of ``build_fn`` to ``model``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -54,6 +54,11 @@ Scikit-Learn on the other hand does not support these arguments, and using them 
 As a compromise, SciKeras supports these arguments, but we recommended that you set parameters using the constructor
 or ``set_params`` for first-class SciKeras support.
 
+.. warning::
+
+   Passing keyword arguments to ``fit`` and ``predict`` is deprecated and will be removed in a future version of SciKeras.
+
+
 For example, to declare ``batch_size`` in the constructor:
 
 .. code:: diff
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -77,8 +77,8 @@ If you want to change the parameters on a live instance, you can do:
    clf.set_params(fit__batch_size=32, predict__batch_size=10000)
    clf.fit(...)
 
-This effectively gives you the same result as passing ``**kwargs``, just with
-one more function call.
+Functionally, this is the same as passing ``**kwargs`` to ``fit``, just with one more function call.
+This is what Scikti-Learn does in the background for hyperparameter tuning.
 
 Renaming of ``build_fn`` to ``model``
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -49,9 +49,15 @@ pass your loss function to the constructor:
 Variable keyword arguments in fit and predict
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-In a future release of SciKeras, variable keyword arguments (commonly referred to as
-``**kwargs``) will be removed from fit and predict. To future
-proof your code, you should instead declare these parameters in your constructor:
+Keras supports a large number of variable keyword arguments keyword arguments (commonly referred to as
+``**kwargs``) for ``fit`` and ``predict``. SciKeras supports these as well,
+but passing arguments in this way means that they will not be hyperparameter-tunable, can't be
+pickled, etc.
+
+In general, it is better to declare these parameters in the constructor of your model.
+While this is not possibel with all parameters, it is not a problem for the vast majority of them.
+
+As an example, we declare ``batch_size`` in the constructor:
 
 .. code:: diff
 
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -77,7 +77,7 @@ If you want to change the parameters on a live instance, you can do:
    clf.set_params(fit__batch_size=32, predict__batch_size=10000)
    clf.fit(...)
 
-Functionally, this is the same as passing ``**kwargs`` to ``fit``, just with one more function call.
+Functionally, this is the same as passing these parameters to ``fit``, just with one more function call.
 This is what Scikti-Learn does in the background for hyperparameter tuning.
 
 Renaming of ``build_fn`` to ``model``
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -1435,7 +1435,7 @@ class KerasClassifier(BaseWrapper):
         Warnings
         --------
             Passing estimator parameters as keyword arguments (aka as ``**kwargs``) to ``predict_proba`` is not supported by the Scikit-Learn API,
-            and may be deprecated in SciKeras.
+            and will be removed in a future version of SciKeras.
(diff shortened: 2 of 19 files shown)


adriangb:
ccing @stsievert

stsievert:
Hm... I think this is a judgement call.

I think having Keras compatibility is a core feature of this library. But this library also tries to bring the Scikit-learn API to Keras models, and the Scikit-learn API says `fit`/etc should only accept data-dependent arguments (not `batch_size`/etc).

If I were developing this library, I think I'd keep an error message but remove all specific version info (and maybe reword the message a bit to warn about the API difference).

adriangb:
Let me try some rewording instead of removal and let's see how it looks.

adriangb:
Here's the new proposed warning:
```
Use of `**kwargs` for `fit` is not fully supported by the sklearn API.  Instead, set fit arguments at initialization (i.e., `BaseWrapper(fit__batch_size=32)`)
```

I also added a `break` so that only 1 kwarg is warned against each time `fit`/`predict` is called.

stsievert:
> I also added a break so that only 1 kwarg is warned against each time fit/predict is called.

Eh, the goal of this warning is to encourage changing behavior. Making the warnings extra annoying for multiple arguments is a good way to do that. Or, you could consolidate the message:

> Only keyword arguments that depend on the data `X` and `y` should be passed to `fit`. The keyword arguments `batch_size`, `verbose` should be set via the parameters `fit__batch_size`/`batch_size`, `fit__verbose`/`verbose`. This can be done via `set_params` or at initialization with `BaseWrapper(..., fit__batch_size=32, fit__verbose=False)`.

adriangb:
Point noted, thanks for the feedback. Let me try another revision.

docs/source/migration.rst
```diff
@@ -70,7 +67,18 @@ Or to declare separate values for ``fit`` and ``predict``:
...
+
+If you want to change the parameters on a live instance, you can do:
+
+.. code:: python
+
+   clf = KerasClassifier(...)
+   clf.set_params(fit__batch_size=32, predict__batch_size=10000)
+   clf.fit(...)
+
+This effectively gives you the same result as passing ``**kwargs``, just with
```
stsievert, line 80:
```suggestion
Functionally, this is the same as passing these parameters during initialization, just with
```

scikeras/wrappers.py
```diff
@@ -35,7 +35,11 @@
 from scikeras.utils.transformers import ClassifierLabelEncoder, RegressorTargetEncoder
 
 
-_kwarg_warn = "Passing `**kwargs` to `{0}` is not fully supported by the sklearn API. Instead, set arguments at initialization (`BaseWrapper({0}__{1}={2})`) or use `set_params`"
+_kwarg_warn = """Passing `**kwargs` to `{0}` is not fully supported by the sklearn API.
```
stsievert, line 38:
```suggestion
_kwarg_warn = """Passing parameters during to `{0}` is not supported by the Scikit-learn API.
```

scikeras/wrappers.py
```diff
@@ -35,7 +35,11 @@
...
+See our docs for more details: https://www.adriangb.com/scikeras/refs/heads/master/migration.html#variable-keyword-arguments-in-fit-and-predict
+The following `**kwargs` were used:
```
stsievert, line 40:
```suggestion
The following parameters were passed to `{0}`:
```

scikeras/wrappers.py
```diff
@@ -706,9 +710,10 @@ def fit(self, X, y, sample_weight=None, **kwargs) -> "BaseWrapper":
             A reference to the instance that can be chain called
             (ex: instance.fit(X,y).transform(X) )
         """
-        for k, v in kwargs.items():
-            warnings.warn(_kwarg_warn.format("fit", k, v))
-            break  # only warn about the first kwarg
+        if kwargs:
+            example_kwarg = next(iter(kwargs.keys()))
+            kwarg_list = "`\n * `".join(kwargs.keys()) + "`"
```
stsievert, line 715: Shouldn't the error message display the value too?

scikeras/wrappers.py
stsievert: And shouldn't the error message provide a method to resolve the warning?

docs/source/migration.rst
adriangb: I am specifically trying to give users a way that they can do this if they do _not_ want to pass the parameters via initialization, hence the comparison to passing `**kwargs`. The first part looks good though.

scikeras/wrappers.py
adriangb: What are the values needed for? I don't see how a developer would use that to fix the issue, they can just look up the value in their code.

scikeras/wrappers.py
adriangb: It links to the docs, which I think provide a more detailed explanation and avenues for resolution than is possible in a warning. Do you feel that linking to the docs does not accomplish this?

scikeras/wrappers.py
adriangb: `Passing parameters during to`?  Is that an extra `to`? Wouldn't this be confused with `fit takes no parameters`?

scikeras/wrappers.py
stsievert: Why not show the values? There's no real downside, especially because you're showing each value on a newline at most initialization parameters are basic types (int, str, float, etc).

scikeras/wrappers.py
stsievert: Yes, linking to the docs provides an example. Yes, users can look at the docs and that will resolve their issue. In fact, you *should* link to the full example where you resolve the issue.

I'm not asking to remove the link. I think an error message should say "to resolve this issue, [method]. For more detail, see [link]."

I think this makes SciKeras easier to use because it requires less work to get the code to run. What if I'm developing without internet access? I've often done that on airplanes or road trips.

scikeras/wrappers.py
stsievert: I meant "parameters" as in "estimator parameters" (e.g., `batch_size`). I've modified the suggestion.

scikeras/wrappers.py
adriangb: That's fair. Do you think we can say `put the parameters in the constructor or use set_params` or something simple like that, without injecting parameter-value examples?

docs/source/migration.rst
stsievert: I'm confused by your statement. I only tried to replace "`**kwargs`" with words. Wouldn't you pass `**kwargs` to `__init__`?

scikeras/wrappers.py
adriangb: > There's no real downside

I think the main downside for long/large warnings is that the information gets lost in the noise. Repeated 20+ line warnings can make it hard to know what is even important.

> most initialization parameters are basic types

You are right, most are basic types. But several, like loss, optimizer, metrics, callbacks, etc. are at the very least iterables/dicts and at the very worst dicts of instances.

docs/source/migration.rst
adriangb: Sorry if it was confusing. What I meant is that this should be the same as passing `**kwargs` to `fit`. So combining my thoughts and yours, I think it would look something like:
```
Functionally, this is the same as passing these parameters to ``fit``, just with one more function call.
```

scikeras/wrappers.py
stsievert: > Repeated 20+ line warnings can make it hard to know what is even important.

Repetition is not an issue. From the [warnings docs][1], "Repetitions of a particular warning for the same source location are typically suppressed." Plus, this warning is only issued once per `fit`/`predict` call (which aren't repeated very often; I would say `partial_fit` is repeated frequently).

I think having one 20+ line warning is okay. I think this warning will be short (<= 4 lines) the vast majority of times it's issued.


[1]:https://docs.python.org/3/library/warnings.html

scikeras/wrappers.py
stsievert: Why the hesitation to include parameter-value examples? I suppose you could use an illustrative example like `foo="bar"`.

scikeras/wrappers.py
adriangb: You make some good points. Besides, we are currently printing these messages with the values, etc, so it's certainly not going to be _worse_ than what we currently have.

I've pushed a commit to print the values, as well as give examples using fixed dummy `foo=bar` kwargs (https://github.com/adriangb/scikeras/pull/198#discussion_r577793274).

scikeras/wrappers.py
```diff
@@ -925,7 +926,6 @@ def predict(self, X, **kwargs):
             and n_features is the number of features.
         **kwargs : Dict[str, Any]
             Extra arguments to route to ``Model.predict``.
-            This functionality has been deprecated, and will be removed in SciKeras 1.0.0.
```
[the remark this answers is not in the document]
stsievert: This goes for all the functions below too.

stsievert (commented):
This looks a lot better.

Style nit: here's some rewording I would do around the warning (more lines, reorganizing words for `**kwargs`):

```
_kwarg_warn = """Passing estimator parameters as keyword arguments to `{0}` (aka as `**kwargs`)  is not supported by the Scikit-Learn API, and will be deprecated in SciKeras.

To resolve this issue, either set these parameters in the constructor (`BaseWrapper(..., foo=bar)`)  or via `set_params` (`est.set_params(foo=bar)`). The following parameters were passed to `{0}`:

{1}

More detail is available at https://www.adriangb.com/scikeras/migration.html#variable-keyword-arguments-in-fit-and-predict
"""
```

scikeras/wrappers.py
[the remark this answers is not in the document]
adriangb: warning section added

adriangb:
> Style nit: here's some rewording I would do around the warning (more lines, reorganizing words for `**kwargs`):

Incorporated for the most part

scikeras/wrappers.py
```diff
@@ -697,20 +704,23 @@ def fit(self, X, y, sample_weight=None, **kwargs) -> "BaseWrapper":
...
         **kwargs : Dict[str, Any]
             Extra arguments to route to ``Model.fit``.
-            This functionality has been deprecated, and will be removed in SciKeras 1.0.0.
-            These parameters can also be specified by prefixing `fit__` to a parameter at initialization;
-            e.g, `BaseWrapper(..., fit__batch_size=32, predict__batch_size=1000)`.
+
+        Warnings
+        --------
+            Passing estimator parameters as keyword arguments (aka as ``**kwargs``) to ``fit`` is not supported by the Scikit-Learn API,
+            and may be deprecated in SciKeras.
```
stsievert, line 711:
```suggestion
            and will be deprecated in SciKeras.
```

A ways down the road, I think this behavior will be deprecated – probably a year or two after TF drops support for Scikit-learn wrappers.

I don't think warnings should be issued for behavior that "may be deprecated."

stsievert (commented):
Small style nit: deprecate is means to be "will be removed."

scikeras/wrappers.py
adriangb: > I suppose you could use an illustrative example like foo="bar"

The warning now has an illustrative example like you suggest above.

scikeras/wrappers.py
adriangb: This suggestion is now outdated, but we have largerly incorporated it, so I am going to resolve this comment.

adriangb:
@stsievert  could you take a look at the unresolved comments? Thanks

docs/source/migration.rst
stsievert: > Functionally, this is the same as passing these parameters to ``fit``, just with one more function call.

Yeah, I think that's a good representation of the message you were trying to convey. I might add a couple words about how it's better because it follows the Sciikit-learn API.

scikeras/wrappers.py
```diff
@@ -36,12 +35,21 @@
 from scikeras.utils.transformers import ClassifierLabelEncoder, RegressorTargetEncoder
 
 
+_kwarg_warn = """Passing estimator parameters as keyword arguments (aka as `**kwargs`) to `{0}` is not supported by the Scikit-Learn API, and will be removed in a future version of SciKeras.
+
+To resolve this issue, either set these parameters in the constructor (`BaseWrapper(..., foo=bar)`) or via `set_params` (`est.set_params(foo=bar)`). The following parameters were passed to `{0}`:
```
stsievert, line 40:
```suggestion
To resolve this issue, either set these parameters in the constructor (e.g., `est = BaseWrapper(..., foo=bar)`) or via `set_params` (e.g., `est.set_params(foo=bar)`). The following parameters were passed to `{0}`:
```

docs/source/migration.rst
adriangb: Edited to:

```
Functionally, this is the same as passing these parameters to ``fit``, just with one more function call.
This is much more compatible with the Scikit-Learn API.
In fact, this is what Scikti-Learn does in the background for hyperparameter tuning.
```

adriangb:
Thanks for this last review @stsievert . I think this is ready.

PR #200, opened 2021-02-19
DOC/ENH: Better AutoEncoder examples & support

adriangb:
- Much cleaner AutoEncoder implementation in docs
- Move model auto-compilation into `_fit` to allow for model initialization without compilation. As far as I can tell, the previous behavior did not hurt anything aside from performing un unnecessary compilation of models that are not meant to be fitted. This not only saves that compilation, but prevents accidentally fitting those models (an error would be raised because they do not have an optimized).

### DIFF
--- a/docs/source/notebooks/AutoEncoders.md
+++ b/docs/source/notebooks/AutoEncoders.md
@@ -83,7 +83,7 @@ We will be defining a very simple autencoder. We define _three_ models:
 
 The only variable we give our model is the encoding dimensions, which will be a hyperparemter of our final transformer.
 
-We will build these models within `_keras-build_fn` so that we can store our encoders and decoders as seperate models and call them from `transform` and `inverse_transform` respectively.
+We will build these models within `_keras_build_fn` so that we can store our encoders and decoders as seperate models and call them from `transform` and `inverse_transform` respectively.
 
 For a background on chaining Functional Models like this, see [All models are callable](https://keras.io/guides/functional_api/#all-models-are-callable-just-like-layers) in the Keras docs.
 
--- a/docs/source/notebooks/AutoEncoders.md
+++ b/docs/source/notebooks/AutoEncoders.md
@@ -134,7 +134,7 @@ class AutoEncoder(BaseWrapper, TransformerMixin):
         self._initialize(X=X, y=X)
         return self
 
-    def fit(self, X, *, sample_weight=None):
+    def fit(self, X, *, sample_weight=None) -> "AutoEncoder":
         super().fit(X=X, y=X, sample_weight=sample_weight)
         # at this point, encoder_model_ and decoder_model_
         # are both "fitted" because they share layers w/ model_
@@ -147,12 +147,12 @@ class AutoEncoder(BaseWrapper, TransformerMixin):
         # always maximizes the score (e.g. in GridSearch)
         return 1 - mean_squared_error(self.predict(X), X)
 
-    def transform(self, X):
-        X = self.feature_encoder_.transform(X)
+    def transform(self, X) -> np.ndarray:
+        X: np.ndarray = self.feature_encoder_.transform(X)
         return self.encoder_model_.predict(X)
 
-    def inverse_transform(self, X_tf):
-        X = self.decoder_model_.predict(X_tf)
+    def inverse_transform(self, X_tf: np.ndarray):
+        X: np.ndarray = self.decoder_model_.predict(X_tf)
         return self.feature_encoder_.inverse_transform(X)
 ```
 
--- a/docs/source/notebooks/AutoEncoders.md
+++ b/docs/source/notebooks/AutoEncoders.md
@@ -78,7 +78,7 @@ print(x_test.shape)
 
 We will be defining a very simple autencoder. We define _three_ model architectures:
 
-1. An encoder: a series of densly connected layers culminating in an "output" layer who's size determines our compression ratio.
+1. An encoder: a series of densly connected layers culminating in an "output" layer that determines the encoding dimensions.
 2. A decoder: takes the output of the encoder as it's input and reconstructs the original data.
 3. An autoencoder: a chain of the encoder and decoder that directly connects them for training purposes.
 
@@ -99,6 +99,9 @@ from scikeras.wrappers import BaseWrapper
 class AutoEncoder(BaseWrapper, TransformerMixin):
     """A class that enables transform and fit_transform.
     """
+
+    encoder_model_: BaseWrapper
+    decoder_model_: BaseWrapper
     
     def _keras_build_fn(self, encoding_dim: int, meta: Dict[str, Any]):
         n_features_in = meta["n_features_in_"]
--- a/docs/source/notebooks/AutoEncoders.md
+++ b/docs/source/notebooks/AutoEncoders.md
@@ -6,7 +6,7 @@ jupyter:
       extension: .md
       format_name: markdown
       format_version: '1.3'
-      jupytext_version: 1.10.1
+      jupytext_version: 1.10.2
   kernelspec:
     display_name: Python 3
     language: python
@@ -164,7 +164,7 @@ autoencoder = AutoEncoder(
     loss="binary_crossentropy",
     encoding_dim=32,
     random_state=0,
-    epochs=15,
+    epochs=5,
     verbose=False,
     optimizer="adam",
 )
@@ -272,7 +272,7 @@ deep = DeepAutoEncoder(
     encoding_dim=32,
     hidden_layer_sizes=[128],
     random_state=0,
-    epochs=15,
+    epochs=5,
     verbose=False,
     optimizer="adam",
 )
--- a/docs/source/notebooks/AutoEncoders.md
+++ b/docs/source/notebooks/AutoEncoders.md
@@ -76,11 +76,11 @@ print(x_test.shape)
 
 ## 3. Define Keras Model
 
-We will be defining a very simple autencoder. We define _three_ models:
+We will be defining a very simple autencoder. We define _three_ model architectures:
 
-1. An encoder
-2. A decoder
-3. An autoencoder (which is just a chain of the encoder and decoder)
+1. An encoder: a series of densly connected layers culminating in an "output" layer who's size determines our compression ratio.
+2. A decoder: takes the output of the encoder as it's input and reconstructs the original data.
+3. An autoencoder: a chain of the encoder and decoder that directly connects them for training purposes.
 
 The only variable we give our model is the encoding dimensions, which will be a hyperparemter of our final transformer.
 
--- a/docs/source/notebooks/AutoEncoders.md
+++ b/docs/source/notebooks/AutoEncoders.md
@@ -120,13 +120,26 @@ class AutoEncoder(BaseWrapper, TransformerMixin):
         self.decoder_model_ = BaseWrapper(decoder_model, predict__verbose=self.predict__verbose)
 
         return autoencoder_model
-
-    def fit(self, X, sample_weight=None):
-        super().fit(X=X, y=X, sample_weight=sample_weight)
+    
+    def _initialize(self, X, y=None):
+        X, _ = super()._initialize(X=X, y=y)
+        # since encoder_model_ and decoder_model_ share layers (and their weights)
+        # X_tf here come from random weights, but we only use it to initialize our models
         X_tf = self.encoder_model_.initialize(X).predict(X)
         self.decoder_model_.initialize(X_tf)
+        return X, X
+
+    def initialize(self, X):
+        self._initialize(X=X, y=X)
         return self
-    
+
+    def fit(self, X, *, sample_weight=None):
+        super().fit(X=X, y=X, sample_weight=sample_weight)
+        # at this point, encoder_model_ and decoder_model_
+        # are both "fitted" because they share layers w/ model_
+        # which is fit in the above call
+        return self
+
     def score(self, X) -> float:
         reconstructed = self.predict(X)
         return binary_accuracy(X, reconstructed).numpy().mean()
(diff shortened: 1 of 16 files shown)


adriangb:
For clarity, the autoencoder code now looks like:

```python
class AutoEncoder(BaseWrapper, TransformerMixin):
    """A class that enables transform and fit_transform.
    """
    
    def _keras_build_fn(self, encoding_dim: int, meta: Dict[str, Any]):
        n_features_in = meta["n_features_in_"]

        encoder_input = keras.Input(shape=(n_features_in,))
        encoder_output = keras.layers.Dense(encoding_dim, activation='relu')(encoder_input)
        encoder_model = keras.Model(encoder_input, encoder_output)

        decoder_input = keras.Input(shape=(encoding_dim,))
        decoder_output = keras.layers.Dense(n_features_in, activation='sigmoid', name="decoder")(decoder_input)
        decoder_model = keras.Model(decoder_input, decoder_output)
        
        autoencoder_input = keras.Input(shape=(n_features_in,))
        encoded_img = encoder_model(autoencoder_input)
        reconstructed_img = decoder_model(encoded_img)

        autoencoder_model = keras.Model(autoencoder_input, reconstructed_img)

        self.encoder_model_ = BaseWrapper(encoder_model, predict__verbose=self.predict__verbose)
        self.decoder_model_ = BaseWrapper(decoder_model, predict__verbose=self.predict__verbose)

        return autoencoder_model

    def fit(self, X, sample_weight=None):
        super().fit(X=X, y=X, sample_weight=sample_weight)
        X_tf = self.encoder_model_.initialize(X).predict(X)
        self.decoder_model_.initialize(X_tf)
        return self
    
    def score(self, X) -> float:
        reconstructed = self.predict(X)
        return binary_accuracy(X, reconstructed).numpy().mean()

    def transform(self, X):
        return self.encoder_model_.predict(X)

    def inverse_transform(self, X_tf):
        X = self.decoder_model_.predict(X_tf)
        return self.feature_encoder_.inverse_transform(X)
```

docs/source/notebooks/AutoEncoders.md
```diff
@@ -73,94 +73,83 @@ print(x_test.shape)
...
-        self.tf_est_ = clone(self.tf_est)
-        self.inv_tf_est_ = clone(self.inv_tf_est)
-        self.tf_est_.set_params(fit_model=self.model_)
-        self.inv_tf_est_.set_params(fit_model=self.model_, encoding_dim=self.encoding_dim)
-        X = self.feature_encoder_.transform(X)
-        self.tf_est_.initialize(X=X)
-        X_tf = self.tf_est_.predict(X=X)
-        self.inv_tf_est_.initialize(X_tf)
+        X_tf = self.encoder_model_.initialize(X).predict(X)
+        self.decoder_model_.initialize(X_tf)
```
stsievert, line 125: Shouldn't this code go add to `super().initialize`, not `super().fit`?

adriangb: I'm pretty sure we want to be calling `super().fit`, but I may be misunderstanding the question.

stsievert: Shouldn't this code be used?

``` python
def initialize(self, X_train):
    super().initialize(X_train)
    X_tf = self.encoder_model_.initialize(X_train).predict(X_train)
    self.decoder_model_.initialize(X_tf)
    return self
```

adriangb: I see, you are right. In fact, the way it is currently written, the encoder is getting `X` before it is modified by `target_encoder_`. This could be fixed by calling `target_encoder_` from `AutoEncoder.fit`, but this would result in an extra `target_encoder_` call which is not good.

But I don't think doing it the way you propose is currently possible. This is because `fit` does not call `initialize` but rather `_initialize` because we want to get back `X` & `y` instead of `self. We would have to do:

```python
class AutoEncoder(BaseWrapper, TransformerMixin):

   def _initialize(self, X, y):
       X, y = super()._initialize(X, y)
       X_tf = self.encoder_model_.initialize(X).predict(X)
      self.decoder_model_.initialize(X_tf)
      return X, y
    
    def fit(...):
      return super().fit(...)  # i.e. we no longer modify fit!
```

adriangb: This could be solved if `initialize` retuned `X, y` instead of `self`, but IMO returning `self` is nicer.

stsievert: It could also be solved if `_initialize` were public, correct?

In https://github.com/adriangb/scikeras/pull/200#issuecomment-782310802's `fit` function, I'm having a hard time seeing why no calls to `encoder_model_.fit` or `decoder_model.fit_` are made.

adriangb: > It could also be solved if _initialize were public, correct?

We have a public `initialize`, it's the same as `_initialize` except for the return arguments (`_initialize` returns `X` & `y`, `initialize` returns `self`).

> I'm having a hard time seeing why no calls to encoder_model_.fit or decoder_model.fit_ are made

`encoder_model_` and `decoder_model_`  share layers (and their weights) with `model_`. When `model_` gets fitted, the weights in `encoder_model_` and `decoder_model_` also get updated. Then `encoder_model_` and `decoder_model_` are used as inference-only models in `transform` and `inverse_transform` respectively. That is, it is completely valid to call `predict` on `encoder_model_` and `decoder_model_` without ever calling `fit`.

docs/source/notebooks/AutoEncoders.md
```diff
@@ -73,99 +75,91 @@ print(x_test.shape)
...
-3. One that creates a model that includes only the decoder portion.
+1. An encoder
+2. A decoder
+3. An autoencoder (which is just a chain of the encoder and decoder)
 
 The only variable we give our model is the encoding dimensions, which will be a hyperparemter of our final transformer.
 
-```python
-from tensorflow import keras
+We will build these models within `_keras-build_fn` so that we can store our encoders and decoders as seperate models and call them from `transform` and `inverse_transform` respectively.
```
stsievert, line 86:
```suggestion
We will build these models within `_keras_build_fn` so that we can store our encoders and decoders as seperate models and call them from `transform` and `inverse_transform` respectively.
```

If `_keras_build_fn` is mentioned in the documentation, should it be private?

docs/source/notebooks/AutoEncoders.md
```diff
@@ -73,99 +75,91 @@ print(x_test.shape)
...
 
-We will be defining a very simple autencoder. We define _three_ model building methods:
+We will be defining a very simple autencoder. We define _three_ models:
 
-1. One to build a full end-to-end autoencoder.
-2. One to create a model that includes only the encoder portion.
-3. One that creates a model that includes only the decoder portion.
+1. An encoder
+2. A decoder
+3. An autoencoder (which is just a chain of the encoder and decoder)
```
stsievert, line 82: These models are of different types. I would add a paragraph explaining that choice.

docs/source/notebooks/AutoEncoders.md
```diff
@@ -73,99 +75,91 @@ print(x_test.shape)
...
-        self.tf_est_.initialize(X=X)
-        X_tf = self.tf_est_.predict(X=X)
-        self.inv_tf_est_.initialize(X_tf)
+        X_tf = self.encoder_model_.initialize(X).predict(X)
+        self.decoder_model_.initialize(X_tf)
         return self
+    
+    def score(self, X) -> float:
+        reconstructed = self.predict(X)
+        return binary_accuracy(X, reconstructed).numpy().mean()
```
stsievert, line 132: What type are `X` and `reconstructed`? It appears that have to be TF arrays.

Why can't you do `return (self.predict(X) == X).mean()`?

adriangb: It could be made public. But generally I think of making a method public when it needs to be called from outside the class. This only ever called from `fit`, so to me it makes sense to keep it private.

docs/source/notebooks/AutoEncoders.md
adriangb: It could be made public. But generally I think of making a method public when it needs to be called from outside the class. This only ever called from fit, so to me it makes sense to keep it private.

docs/source/notebooks/AutoEncoders.md
adriangb: Wrong thread for the previous comment.

As far as I can tell, they are Numpy arrays. TF metrics/losses usually also accept Numpy inputs.

> Why can't you do return (self.predict(X) == X).mean()?

More than likely that we can do that, I was probably just being a bonehead.

docs/source/notebooks/AutoEncoders.md
stsievert: What determines if a function needs to be called? Why doesn't this usage qualify?

My metric for public/private "will users every call this function?" I think if a function is shown in an example on the documentation page, then that function should be public.

docs/source/notebooks/AutoEncoders.md
stsievert: It'd be really useful to add type hints. I asked because calling `.numpy` on a NumPy array raises an exception.

stsievert: See #201.

docs/source/notebooks/AutoEncoders.md
adriangb: What do you mean by `different types`? Different architectures?

docs/source/notebooks/AutoEncoders.md
adriangb: Thanks for opening #201.

> Why can't you do return (self.predict(X) == X).mean()?

We could. I made a dumb mistake though and put in `binary_accuracy`, I meant to use `binary_crossentropy`. But really it would be better to use `MSE`, since that is an actual metric. I pushed a commit to do just that. Well, actually we use `1-MSE` because sklearn always maximizes `score` in `GridSearchCV` and such. I added a comment to to clarify that in the code.

docs/source/notebooks/AutoEncoders.md
adriangb: I do think there's good arguments for making it public. I opened #202 to track this. I don't think this should block this PR. `_keras_build_fn` is mentioned and used elsewhere in the docs, so we can't resolve this here.

docs/source/notebooks/AutoEncoders.md
adriangb: I think the type-hint can be continued in #201.
For this PR, I added some type hints in the notebook for the return type from `Model.predict`. That that coupled with the removal of the code that originally caused the confusion should be enough for this PR.

docs/source/notebooks/AutoEncoders.md
stsievert: The autoencoder is a `keras.Model`. The encoder and decoder are `BaseWrapper`s.

adriangb: I see what you mean now, sorry for being slow.

The intention of this section was to discuss the Keras Model architectures. I edited the text to reflect this. I also added class-attribute type-hints for `encoder_model_` and `decoder_model_` indicating that they are instances of `BaseWrapper`. I hope this helps clarify things.

docs/source/notebooks/AutoEncoders.md
adriangb: Regarding the use of `_initialize`, the way it stands now the `AutoEncoder` class is a subclass of `BaseWrapper` which is overriding the protected member method `BaseWrapper._initialize` which is totally fine, at least in other languages where these types of things are actually enforced. Some better documentation of `_initialize` might help understand why and how developers subclassing `BaseWrapper` might want to call it, but I think that is beyond the scope of this PR.

docs/source/notebooks/AutoEncoders.md
stsievert: If correct, maybe add a line saying something similar to the following?

> The encoder and decoder are views to the first/last layers of the autoencoder model. They'll be directly used in `predict`/`inverse_transform`/etc, so let's create some SciKeras models with those layers in `{encoder, decoder}_model_`.

adriangb: I added a similar sentence. Thank you for the suggestion.

docs/source/notebooks/AutoEncoders.md
```diff
@@ -73,99 +76,109 @@ print(x_test.shape)
...
+1. An encoder: a series of densly connected layers culminating in an "output" layer that determines the encoding dimensions.
+2. A decoder: takes the output of the encoder as it's input and reconstructs the original data.
+3. An autoencoder: a chain of the encoder and decoder that directly connects them for training purposes.
 
 The only variable we give our model is the encoding dimensions, which will be a hyperparemter of our final transformer.
 
-```python
-from tensorflow import keras
+The encoder and decoder are views to the first/last layers of the autoencoder model. They'll be directly used in predict/inverse_transform/etc, so let's create some SciKeras models with those layers in {encoder, decoder}_model_.
+We will build these models within `_keras-build_fn` so that we can store our encoders and decoders as seperate models and call them from `transform` and `inverse_transform` respectively.
```
stsievert, line 88:
```suggestion
We will build these models within `_keras-build_fn` so that we can store our encoders and decoders as seperate models and call them from `transform` and `inverse_transform` respectively. 
```
```suggestion
We will build these models within `_keras_build_fn` so that we can store our encoders and decoders as separate models and call them from `transform` and `inverse_transform` respectively. 
```

adriangb: Sorry, I had un-pushed changes, apologies for the crosstalk. The wording now reads:

```
The encoder and decoder are views to the first/last layers of the autoencoder model.
They'll be directly used in ``transform`` and ``inverse_transform``, so we'll create some SciKeras models with those layers
and save them as in `encoder_model_` and `decoder_model_`. All three models are created within `_keras_build_fn`.
```

I think this is pretty consistent with your previous suggestion? Sorry again for forgetting to push that.

stsievert: Yeah, that all looks good

adriangb:
Thank you for all of your feedback @stsievert! I think this is pretty ready, I'm going to merge it.

PR #203, opened 2021-02-20
MAINT: add type-hints to _saving_utils.py

adriangb:
- Add type hints
- Use tarfile instead of zipfile (we just want to create an archive)

### DIFF
--- a/scikeras/_saving_utils.py
+++ b/scikeras/_saving_utils.py
@@ -1,19 +1,20 @@
 import os
+import tarfile
 import tempfile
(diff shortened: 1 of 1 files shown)


adriangb:
Benchmarking this against `master` and against a pure `SaveModel` as a control.

Unsurprisingly, simply taring the files instead of zipping (and compressing) each one gives better performance, to the point where there is not measurable overhead over `SaveModel`.

**master**
download (2)

**this branch**

download (3)

Issue #206, opened 2021-02-24
No default loss for KerasClassifier


stsievert:
Why does KerasClassifier not have a default loss?

https://github.com/adriangb/scikeras/blob/d83bffaf8497289d9755f42654bdcc15584e68e6/scikeras/wrappers.py#L1229-L1231

Shouldn't there be a least a sensible default that works in most use cases?

**Reference issues/PRs**
https://github.com/dask/dask-ml/pull/794

adriangb:
Keras itself does not provide a default loss. The loss depends on the dataset, so I don't think it makes sense to provide a concrete loss as a default. At most, we could accept a sting like `"auto"` or `"infer"` that could be passed to `target_encoder` which can then determine an appropriate loss for the specific dataset.

But this also sounds like something users could implement, or that could be shown in notebooks, I'm not sure it should be a part of the core of SciKeras.

stsievert:
Why should the user specify the `loss` parameter for their classifier to work? Why is it important to expose the loss function to the user?

Aren't particular classes of loss function pretty commonly used for classification regardless of dataset? (like cross entropy?).

How does the Scikit-learn API choose the loss for their classifier when the loss function can be varied? (e.g., [SGDClassifier][1]).

[1]:https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier

adriangb:
> Why is it important to expose the loss function to the user?

Because the loss function depends not only on the input data (which is not to hard to parse, especially since Scikit-Learn has tools like `type_of_target`) but also on the Model architecture, which is considerably harder to parse and bin into clear choices. I believe this is why Keras does not provide a default loss function.

> How does the Scikit-learn API choose the loss for their classifier when the loss function can be varied? 

Scikit-Learn has full control of the internals. They can choose/build a model to match a loss function; Scikeras/Keras have to work with whatever Model users create. That said, I don't really know what Scikit-Learn does internally, so I could be missing something.

I don't mean to be overly negative on the idea. I do think that abstracting things like the loss function from users is good. The simpler it is for users to get started, the better.  But I would like to work on core parts of SciKeras (like #167) that enable Scikit-Learn/Keras interoperability first before adding more convenience layers.

stsievert:
To be clear, `BaseWrapper` should absolutely not perform any heuristic processing. It should leave everything untouched. If a user wants to do anything not-standard, that's the class they should use.

[...]

Issue #207, opened 2021-02-24
Sequential model doesn't have outputs


stsievert:
Shouldn't this code work?

``` python
from sklearn.datasets import make_classification
from scikeras.wrappers import KerasClassifier
import tensorflow as tf

def model():
    model = tf.keras.Sequential()
    model.add(tf.keras.layers.Dense(8))
    model.add(tf.keras.layers.Dense(1))
    return model

X, y = make_classification(n_features=8)
est = KerasClassifier(model=model, loss="sparse_categorical_crossentropy")
est.fit(X, y=y)
```

This throws a `ValueError: object of type `NoneType` [self.model_.outputs] has no len()`.

<details>
<summary>Full traceback</summary>

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/Downloads/_junk2.py in <module>
     11 X, y = make_classification(n_features=8)
     12 est = KerasClassifier(model=model, loss="sparse_categorical_crossentropy")
---> 13 est.fit(X, y=y)

~/anaconda3/envs/scikeras/lib/python3.7/site-packages/scikeras/wrappers.py in fit(self, X, y, sample_weight, **kwargs)
   1375             sample_weight = 1 if sample_weight is None else sample_weight
   1376             sample_weight *= compute_sample_weight(class_weight=self.class_weight, y=y)
-> 1377         super().fit(X=X, y=y, sample_weight=sample_weight, **kwargs)
   1378         return self
   1379

~/anaconda3/envs/scikeras/lib/python3.7/site-packages/scikeras/wrappers.py in fit(self, X, y, sample_weight, **kwargs)
    739             epochs=getattr(self, "fit__epochs", self.epochs),
    740             initial_epoch=0,
--> 741             **kwargs,
    742         )
    743

~/anaconda3/envs/scikeras/lib/python3.7/site-packages/scikeras/wrappers.py in _fit(self, X, y, sample_weight, warm_start, epochs, initial_epoch, **kwargs)
    855         X = self.feature_encoder_.transform(X)
    856
--> 857         self._check_model_compatibility(y)
    858
    859         self._fit_keras_model(

~/anaconda3/envs/scikeras/lib/python3.7/site-packages/scikeras/wrappers.py in _check_model_compatibility(self, y)
    541             # we recognize the attribute but do not force it to be
    542             # generated
--> 543             if self.n_outputs_expected_ != len(self.model_.outputs):
    544                 raise ValueError(
[...]

PR #208, opened 2021-02-27
ENH: Add default loss to KerasClassifier

stsievert:
It adds `loss="categorical_crossentropy"` to KerasClassifier by default. It adds some protection in case the user passes multiple classes in `y` but the model is only configured for one class.

**Reference issues/PR**
This PR closes https://github.com/adriangb/scikeras/issues/206

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -1327,12 +1327,12 @@ class KerasClassifier(BaseWrapper):
             ):
                 raise ValueError(
                     "The model is configured to have one output, but the "
-                    f"loss='{self.loss}' can not handle multiple outputs. "
-                    "This loss is often used with one-hot encoded targets."
+                    f"loss='{self.loss}' is expecting multiple outputs "
+                    "(which is often used with one-hot encoded targets). "
                     "More detail on Keras losses: https://keras.io/api/losses/"
                 ) from e
             else:
-                raise ValueError("{self.loss}, {self.model_.outputs}")
+                raise e
 
 
     @staticmethod
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -1142,12 +1142,18 @@ class KerasClassifier(BaseWrapper):
         an instance of tf.keras.optimizers.Optimizer
         or a class inheriting from tf.keras.optimizers.Optimizer.
         Only strings and classes support parameter routing.
-    loss : Union[Union[str, tf.keras.losses.Loss, Type[tf.keras.losses.Loss], Callable], None], default None
+    loss : Union[Union[str, tf.keras.losses.Loss, Type[tf.keras.losses.Loss], Callable], None], default "categorical_crossentropy"
         The loss function to use for training.
         This can be a string for Keras' built in losses,
         an instance of tf.keras.losses.Loss
         or a class inheriting from tf.keras.losses.Loss .
         Only strings and classes support parameter routing.
+
+        For convience, the loss defaults to
+        `"categorical_crossentropy"`. This assumes that the model has
+        ``N`` outputs if the dataset has ``N`` classes. It assumes that
+        the input
+
     random_state : Union[int, np.random.RandomState, None], default None
         Set the Tensorflow random number generators to a
         reproducible deterministic state using this seed.
@@ -1257,7 +1263,7 @@ class KerasClassifier(BaseWrapper):
         ] = "rmsprop",
         loss: Union[
             Union[str, tf.keras.losses.Loss, Type[tf.keras.losses.Loss], Callable], None
-        ] = None,
+        ] = "categorical_crossentropy",
         metrics: Union[
             List[
                 Union[
@@ -1309,6 +1315,25 @@ class KerasClassifier(BaseWrapper):
             # check that this is not a multiclass problem missing categories
             target_type = type_of_target(self.classes_)
         return target_type
+
+    def _fit_keras_model(self, *args, **kwargs):
+        try:
+            super()._fit_keras_model(*args, **kwargs)
+        except ValueError as e:
+            if (
+                self.loss == "categorical_crossentropy"
+                and hasattr(self, "model_")
+                and 1 in {o.shape[1] for o in getattr(self.model_, "outputs", [])}
+            ):
+                raise ValueError(
+                    "The model is configured to have one output, but the "
+                    f"loss='{self.loss}' can not handle multiple outputs. "
+                    "This loss is often used with one-hot encoded targets."
+                    "More detail on Keras losses: https://keras.io/api/losses/"
+                ) from e
+            else:
+                raise ValueError("{self.loss}, {self.model_.outputs}")
+
 
     @staticmethod
     def scorer(y_true, y_pred, **kwargs) -> float:
--- a/tests/test_simple_usage.py
+++ b/tests/test_simple_usage.py
@@ -1,0 +1,42 @@
+import numpy as np
+import pytest
+import tensorflow as tf
+from sklearn.datasets import make_classification
+
+from scikeras.wrappers import KerasClassifier
+
+N_CLASSES = 4
+FEATURES = 8
+n_eg = 100
+X = np.random.uniform(size=(n_eg, FEATURES))
+y = np.random.choice(N_CLASSES, size=n_eg)
+
+
+def clf(single_output=False):
+    model = tf.keras.Sequential()
+    model.add(tf.keras.layers.Input(shape=(FEATURES,)))
+    model.add(tf.keras.layers.Dense(FEATURES))
+
+    if single_output:
+        model.add(tf.keras.layers.Dense(1))
+    else:
+        model.add(tf.keras.layers.Dense(N_CLASSES))
+
+    return model
+
+
+def test_classifier_loss_defaults():
+    est = KerasClassifier(model=clf)
+    est.partial_fit(X, y=y)
+    assert est.current_epoch == 1
+
+
+def test_classifier_raises_for_single_output():
+    est = KerasClassifier(model=clf, model__single_output=True)
+    msg = (
+        "one output, but the loss='categorical_crossentropy' "
+        "can not handle multiple outputs"
+    )
+    with pytest.raises(ValueError, match=msg):
+        est.partial_fit(X, y)
+    assert est.current_epoch == 0
--- a/tests/test_simple_usage.py
+++ b/tests/test_simple_usage.py
@@ -8,8 +8,8 @@ from scikeras.wrappers import KerasClassifier
 N_CLASSES = 4
 FEATURES = 8
 n_eg = 100
-X = np.random.uniform(size=(n_eg, FEATURES))
-y = np.random.choice(N_CLASSES, size=n_eg)
+X = np.random.uniform(size=(n_eg, FEATURES)).astype("float32")
+y = np.random.choice(N_CLASSES, size=n_eg).astype(int)
 
 
 def clf(single_output=False):
@@ -25,17 +25,40 @@ def clf(single_output=False):
     return model
 
 
-def test_classifier_loss_defaults():
+def test_classifier_only_model_specified():
+    """
+    This tests uses cases where KerasClassifier works with the default loss.
+    It works for the following cases:
+
+    * binary classification
+    * one hot classification
+    * single class classification
+
+    """
     est = KerasClassifier(model=clf)
     est.partial_fit(X, y=y)
     assert est.current_epoch == 1
 
+    for y2 in [
+        np.random.choice(2, size=len(X)).astype(int),
+        (np.random.choice(2, size=len(X)).astype(int) * 2 - 1),
+        np.ones(len(X)).astype(int),
+        np.zeros(len(X)).astype(int),
+    ]:
+        est = KerasClassifier(model=clf, model__single_output=True)
+        est.partial_fit(X, y=y2)
+        assert est.current_epoch == 1
 
-def test_classifier_raises_for_single_output():
+
+def test_classifier_raises_for_single_output_with_multiple_classes():
+    """
+    KerasClassifier does not work with one output and multiple classes
+    in the target (duh).
+    """
     est = KerasClassifier(model=clf, model__single_output=True)
     msg = (
-        "one output, but the loss='categorical_crossentropy' "
-        "can not handle multiple outputs"
+        "The model is configured to have one output, but the "
+        "loss='categorical_crossentropy' is expecting multiple outputs "
     )
     with pytest.raises(ValueError, match=msg):
         est.partial_fit(X, y)


adriangb (commented):
Thank you for the PR. I left some more general questions.

tests/test_simple_usage.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def test_classifier_only_model_specified():
+    """
+    This tests uses cases where KerasClassifier works with the default loss.
+    It works for the following cases:
+
+    * binary classification
+    * one hot classification
+    * single class classification
```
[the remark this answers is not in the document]
adriangb: I think I see now. This should work!

scikeras/wrappers.py
```diff
@@ -1310,6 +1316,24 @@ def _type_of_target(self, y: np.ndarray) -> str:
...
+                and 1 in {o.shape[1] for o in getattr(self.model_, "outputs", [])}
+            ):
+                raise ValueError(
+                    "The model is configured to have one output, but the "
+                    f"loss='{self.loss}' is expecting multiple outputs "
+                    "(which is often used with one-hot encoded targets). "
+                    "More detail on Keras losses: https://keras.io/api/losses/"
+                ) from e
+            else:
+                raise e
```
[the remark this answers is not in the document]
stsievert: This error message only provides marginal utility: it protects against cases when the model has one output but there are multiple classes.

It can not go in `_check_model_compatibility`; I wait for an error to be raised before issuing this warning (otherwise a model with a single output raises an error).

tests/test_simple_usage.py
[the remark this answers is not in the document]
stsievert: I've made the uses cases where this works more clear in 9358d6e. It works for all major use cases.

This PR simply changes the default loss; it doesn't change compatibility in any way.

scikeras/wrappers.py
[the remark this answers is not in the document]
adriangb: Got it. Is there a specific error message we can check for, like `if "some Keras error" in str(e)`?

> `getattr(self.model_, "outputs", [])`

Is this necessary? `model_` should always have an `outputs` attribute, except in the case described in #207, but that should be a separate check/error.

> f"loss='{self.loss}' is expecting multiple outputs "

Can you clarify what you mean by a loss expecting a number of outputs? My understanding is that Keras "broadcasts" losses to outputs, so if you give it a scalar loss (ie.. `loss="bce"`) with 2 outputs (i.e. `len(model_.outputs) == 2`), it will implicitly compile the model with `loss=[original_loss] * len(outputs)`. But you can actually map losses to outputs manually, by passing `loss=["bce", "mse"]` or `loss={"out1": "bce", "out2": "mse"}`. From the tests, it seems like by "loss is expecting multiple outputs" you mean that there is a single output unit but multiple classes, which I feel like could be confused with the above concept of configuring a loss for multiple outputs.

I'm also curious about the iteration through outputs (`1 in {o.shape[1] for o in self.model_.outputs}`). SciKeras does not support >1 output out of the box (users need to override `target_encoder`) so it seems a bit strange to try to account for that when using the default loss. I feel that using the default loss should only be supported for the simple single-output cases that `target_encoder` supports.

adriangb:
Should we do anything for models with a single output and a single linear output unit and 2 classes? This is a subset of your error catching / test, but Keras won't raise a ValueError for it.  Instead, it will train and not learn anything. The same model with `loss="bce"` will easily predict with 100% accuracy:

```python
import numpy as np
import tensorflow as tf

N_CLASSES = 2
FEATURES = 1
n_eg = 10000
y = np.random.choice(N_CLASSES, size=n_eg).astype(int)
X = y.reshape(-1, 1).astype("float32")


def clf():
    model = tf.keras.Sequential()
    model.add(tf.keras.layers.Input(shape=(FEATURES,)))
    model.add(tf.keras.layers.Dense(4))
    model.add(tf.keras.layers.Dense(1))
    return model


model = KerasClassifier(clf, loss="categorical_crossentropy", epochs=50)

model.fit(X, y).score(X, y)  # ~ 0.5
```

scikeras/wrappers.py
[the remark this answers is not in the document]
adriangb: As a side note: I think giving users better errors and validating their inputs like you are doing here can be a very valuable part of SciKeras, but currently it is done in an ad-hoc manner via `_check_model_compatibility`, etc. I think if we add more of these types of things, it would be nice to have an organized interface for it. I opened #209 to try to brainstorm ideas for this.

scikeras/_types.py
```diff
@@ -0,0 +1,18 @@
...
+from tensorflow.keras.metrics import Metric as TF_Metric
+from tensorflow.keras.optimizers import Optimizer as TF_Optimizer
+
+
+Model = Union[Callable[..., keras.Model], keras.Model]
+RandomState = Union[int, np.random.RandomState]
+Optimizer = Union[str, TF_Optimizer, Type[TF_Optimizer]]
+Loss = Union[str, TF_Loss, Type[TF_Loss], Callable]
+Metrics = Union[List[Union[str, TF_Metric, Type[TF_Metric], Callable]]]
+Callbacks = Union[List[Union[TF_Callback, Type[TF_Callback]]]]
```
adriangb, line 18: This is great @stsievert!

adriangb (commented):
Looks like you're making some progress, thank you for keeping at it. If you're budding up against annoying / hard to debug tests let me know, I'm happy to jump in and try to iron it out.

stsievert:
I've got most of the issues resolved:

* A warning is issued for a binary classification target but `loss="categorical_crossentropy"`; the user might have compiled their own model.

The tests on fully compliant Scikit-learn estimators are failing now. I'd appreciate some help.

scikeras/wrappers.py
```diff
@@ -1142,12 +1124,26 @@ class KerasClassifier(BaseWrapper):
...
+    loss : Optional[Union[str, tf.keras.losses.Loss, Type[tf.keras.losses.Loss], Callable]], default "categorical_crossentropy"
         The loss function to use for training.
         This can be a string for Keras' built in losses,
         an instance of tf.keras.losses.Loss
         or a class inheriting from tf.keras.losses.Loss .
         Only strings and classes support parameter routing.
+
+        For convience, the loss defaults to
+        `"categorical_crossentropy"`. This assumes that the model has
+        ``N`` outputs if the dataset has ``N`` classes. It works with
```
adriangb, line 1136:
```suggestion
        1 output with ``N`` output units if the dataset has ``N`` classes. It works with
```

tests/test_simple_usage.py
```diff
@@ -0,0 +1,103 @@
...
+    else:
+        raise ValueError("use_case={use_case} not recognized")
+
+    est = KerasClassifier(model=shallow_net, model__single_output=model__single_output)
+    if "binary" in use_case:
+        with pytest.warns(UserWarning, match="Set loss='binary_crossentropy'"):
+            est.partial_fit(X, y)
+        est.set_params(loss="binary_crossentropy")
+
+    est.partial_fit(X, y=y)
```
adriangb, line 63:
```suggestion
    with pytest.warns(None) as record:
        est.partial_fit(X, y=y)
        assert not any("Set loss='binary_crossentropy'") in str(r.message) for r in record)
```

tests/test_simple_usage.py
```diff
@@ -0,0 +1,103 @@
...
+def test_classifier_raises_for_single_output_with_multiple_classes():
+    """
+    KerasClassifier does not work with one output and multiple classes
+    in the target (duh).
+    """
+    est = KerasClassifier(model=shallow_net, model__single_output=True)
+    y = np.random.choice(N_CLASSES, size=len(X))
+    msg = (
+        "The model is configured to have one output neuron, but the "
+        "loss='categorical_crossentropy' is expecting multiple outputs "
```
adriangb, line 76:
```suggestion
       f"loss='categorical_crossentropy' with {N_CLASSES} classes requires {N_CLASSES} output neurons "
```

tests/test_simple_usage.py
```diff
@@ -0,0 +1,103 @@
...
+    est.partial_fit(X, y)
+    assert est.score(X, y) >= 0.9
+
+
+def test_regressor_default_loss():
+    y = np.random.uniform(size=len(X))
+    est = KerasRegressor(model=shallow_net, model__single_output=True)
+    assert est.loss == "mse"
+    est.partial_fit(X, y)
+    assert est.model_.loss.__name__ == "mean_squared_error"
```
adriangb, line 103: Is the assertion that the "long" name for the loss was used in the model necessary here? I don't see the same assertion for classifiers.

adriangb (commented):
> A warning is issued for a binary classification target but loss="categorical_crossentropy"; the user might have compiled their own model.

👍 

> I removed the GitHub action dependence; why should the tests depend on linting? It's really annoying.

This is a somewhat arbitrary choice. I'm open to removing it, but I'd like that done in a separate PR. If you use the git pre-commit hook, you should never have a problem being annoyed by linting because you won't even be able to make a commit that fails linting (unless you manually override it).

> The tests on fully compliant Scikit-learn estimators are failing now

I'll take a look.

scikeras/wrappers.py
```diff
@@ -1636,7 +1679,7 @@ def scorer(y_true, y_pred, **kwargs) -> float:
 
     def _validate_data(
         self, X=None, y=None, reset: bool = False, y_numeric: bool = False
-    ) -> Tuple[np.ndarray, Union[np.ndarray, None]]:
+    ) -> Tuple[np.ndarray, Optional[np.ndarray]]:
```
adriangb, line 1682: I don't think this should be changed from `Union[..., None]` -> `Optional[...]`. Yes, they are technically equivalent, but semantically `Optional` does not make sense if there is no default argument.

scikeras/_types.py
adriangb, line 18: I think this file should be renamed to ` py.typed` for future MyPy compatibility, etc. But not 100% sure on that.

adriangb:
At least some of the problems seem to boil down to the fact that some tests were doing something like this:

```python
def dynamic_classifier(..., compile_kwargs):
    loss = compile_kwargs["loss"] or ...  # behavior changes if instead of None it gets a functions/string
```

We can certainly refactor the tests to try and improve on this. But looking at these implementations also makes me think that we may want to make the default `loss="auto"`, which makes it possible to know if users passed an explicit loss or used the default loss. Then, we could change our tests to be:

```python
class DynamicClassifier(...):
    def _keras_build_fn(self, compile_kwargs):
        if self.loss == "auto":
            loss = None  # dynamically determined in conjunction with the model
        else:
            loss = compile_kwargs["loss"]  # user provided; override auto determination of loss
        ...
```

For converting from `"auto"` -> `"categorical_crossentropy"`
```python
class BaseWrapper(...):
    def _substitute_default_loss(self, loss):
        return loss

    def _get_compile_kwargs(self):
        ...
        comple_kwargs["loss"] = self._substitute_default_loss(comple_kwargs["loss"])
        ...

class KerasClassifier(...):
    def _substitute_default_loss(self, loss):
        if loss == "auto":
            return "categorical_crossentropy"
       return super()._infer_loss(loss)
```

I'm not a huge fan of `_substitute_default_loss`, but perhaps we can come up with a better way.

What do you think of this `loss="auto"` scheme @stsievert ?

tests/mlp_models.py
```diff
@@ -27,10 +27,10 @@ def dynamic_classifier(
         hidden = Dense(layer_size, activation="relu")(hidden)
 
     if target_type_ == "binary":
-        compile_kwargs["loss"] = compile_kwargs["loss"] or "binary_crossentropy"
+        compile_kwargs["loss"] = "binary_crossentropy"
```
adriangb, line 30: Unfortunately it's not that simple... The original intention was "use the inferred loss unless it is explicitly overriden", and there are indeed tests that explicitly override it to test various things (wrong loss, no loss, etc.)

scikeras/_types.py
stsievert: It looks like the file only needs to be created: https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

stsievert: ...but I'm not sure how to instruct PyPI/etc how to install that file in pyproject.toml.

adriangb: Thank for checking. It looks like [this](https://github.com/python-poetry/poetry/issues/1338#issuecomment-703737178) is how. But don't worry about it, I can take care of that in a separate PR.

tests/test_simple_usage.py
stsievert: This assert statement is present to make sure the `BaseWrapper.loss` is mirror in `BaseWrapper.model_.loss`. I'll add a test for `KerasClassifier` too.

adriangb: Maybe this is good use case for `scikeras.utils.loss_name`?

stsievert:
Tests are passing now; I manually specified `loss=None`. I think `loss="auto"` should decide between `binary_crossentropy` and `categorical_crossentropy`; I'm not a huge fan of changing the package implementation based on the tests.

On the whole, this PR makes SciKeras more usable, though technically backwards incompatible for a very narrow use case. That relies on a couple points:

* Simple usage works now. 
    * Before: For some tests in `test_simple_usage.py`, the `master` branch raises a "ValueError: No valid loss function found. You must provide a loss function to train" for most of the tests in `test_simple_usage.py`.
    * After: If I specify a commonly used model (e.g., a regressor with one output node or classifier with `N_CLASSES` output nodes) and supply training data, my Keras model will learn.
    * Technically, this is backward incompatible is compiling their own model and depending on `loss is None`.
* One less error raised:
    * Before: An exception is raised if `BaseWrapper.loss` and `BaseWrapper.model_.loss` are different.
    * After: a warning is issued for the same case. (if the user is compiling their own model, I presume they have a special use case or know what they're doing).

adriangb:
I believe this PR is not breaking basic wrapper backwards compatibility:

```python
import numpy as np
from tensorflow import keras

num_classes = 3
n_features = 4
n_samples = 100

X = np.random.random(size=(n_samples, n_features))
y = np.random.randint(low=0, high=num_classes, size=(n_samples,))

def get_model():
    model = keras.Sequential(
        [
            keras.Input((n_features,)),
            keras.layers.Dense(n_features),
            keras.layers.Dense(num_classes, activation="softmax"),
        ]
    )
    model.compile(loss="sparse_categorical_crossentropy")
    return model

# Works via Keras
get_model().fit(X, y)

# Works via Keras wrappers
keras.wrappers.scikit_learn.KerasClassifier(get_model).fit(X, y)

# Does not work with SciKeras stsievert:clf-default-loss but does work on SciKeras master
KerasClassifier(get_model).fit(X, y)
```

This can be "fixed" by using `KerasClassifier(model=get_model, loss=None).fit(X, y)` like you do in the tests, but this is a breaking change, I'd like to call it out explicitly before moving forward

stsievert:
This error only happens when both of the following occur:

* When the user compiles the model themselves w/ `loss="sparse_categorical_crossentropy"`.
* When the default loss is not changed from `categorical_crossentropy`.

I'm going to raise an exception if the Keras model specifies `loss="sparse_categorical_crossentropy"` and, in the error message, tell the user to how to clear the error (changing the loss parameter in `KerasClassifier.fit` is not allowed by the Scikit-learn API).

adriangb:
Can't this also be solved by just defaulting to `"auto"`?

This also makes me wonder if the design decision to put the description of the data and encoding / decoding into the same place was flawed. Or if perhaps inspecting the compiled model was better than relying on the user's input.

stsievert:
Yes, `loss="auto"` is a good idea. It simplifies the tests, and also allows choosing `loss="bce"` if there is one output node, there's only two unique values in `y` and the user has not compiled the model.

adriangb:
> Yes, `loss="auto"` is a good idea.

I'll try to cook up an implementation of that in the next couple of days. Regardless, thank you for all of the great work you've done on this PR so far, I think it's done a lot to illuminate our options and challenges with this feature.

stsievert:
I'd like to see a new function inserted after model creation and before model compiling to choose the some compile arguments based on the model architecture. It'd be useful to have this function to choose the loss function from the number of output neurons. That'd help a lot with the `loss="auto"` implementation and choosing between `loss="bce"` and `loss="cce"`.

adriangb:
We have `_compile_model`, although it currently also decides if it _should_ compile the model or not. Perhaps we can split that into `_should_compile_model` and `_compile_model` where `_compile_model` is a designated dependency injection point for insertion of arbitrary compile logic based upon model architecture and any other pre-determined parameters (like the output from `target_encoder`)? Then SciKeras can by default use this to handle `loss="auto"`, but users can also override it to support other uses.

stsievert:
I'm thinking of this pseudo-code:

``` python
assert len(self.model_.outputs) == 1
out = self.model_.outputs[0]

if self.loss == "auto":
    if out.shape[1] == 1 and y_n_unique <= 2:
        loss = "binary_crossentropy"
    elif out.shape[1] > 1 and len(y_encoded[0]) == 1:
        loss = "sparse_categorical_crossentropy"
    elif out.shape[1] > 1 and len(y_encoded[0] > 1:
        loss = "categorical_crossentropy"

self._compile_model(loss=loss)
```

This requires that `_build_fn` be called (for `self.model_`) and that the data be encoded (for metadata about `y_encoded`). It looks like this code could go an overwrite of `_ensure_compiled_model` (which needs a better name). `KerasClassifier` would also have to override `_get_compile_kwargs` to return a default loss value (the promise of `compile_kwargs` is that they're all valid arguments to `self.model_.compile`, right?).

adriangb:
This seems pretty similar to what I am proposing in #210 , please check that PR if you can.

adriangb:
> * I removed the GitHub action dependence; why should the tests depend on linting? It's really annoying.

Resolved via #217

PR #210, opened 2021-03-04
attempt at default losses

adriangb:
@stsievert  a parallel proposal to #208 

This implements a default loss `"auto"` for KerasClassifier and KerasRegressor. An appropriate loss function is only selected if:
1. The user did not provide a loss function (i.e. the default "auto" was passed).
2. The user did not compile the model.

For KerasRegressor, it always defaults to "mse" and supports any number of outputs.

For KerasClassifier, it defaults to "binary_crossentropy" for binary targets, "sparse_categorical_crossentropy" for "multiclass" targets and and "categorical_crossentropy" for all other problems. Only single outputs are supported. An error is raised if there is >1 output or if a different task type not listed above is passed (eg: multilabel-indicator).

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -22,6 +22,7 @@ from tensorflow.keras import metrics as metrics_module
 from tensorflow.keras import optimizers as optimizers_module
 from tensorflow.keras.models import Model
 from tensorflow.keras.utils import register_keras_serializable
+from tensorflow.python.types.core import Value
 
 from scikeras._utils import (
     TFRandomState,
@@ -345,27 +346,36 @@ class BaseWrapper(BaseEstimator):
         compile_kwargs = route_params(
             init_params, destination="compile", pass_filter=self._compile_kwargs,
         )
-        compile_kwargs["optimizer"] = _class_from_strings(
-            compile_kwargs["optimizer"], optimizers_module.get
-        )
+        try:
+            compile_kwargs["optimizer"] = _class_from_strings(
+                compile_kwargs["optimizer"], optimizers_module.get
+            )
+        except ValueError:
+            pass  # unknown optimizer
         compile_kwargs["optimizer"] = unflatten_params(
             items=compile_kwargs["optimizer"],
             params=route_params(
                 init_params, destination="optimizer", pass_filter=set(), strict=True,
             ),
         )
-        compile_kwargs["loss"] = _class_from_strings(
-            compile_kwargs["loss"], losses_module.get
-        )
+        try:
+            compile_kwargs["loss"] = _class_from_strings(
+                compile_kwargs["loss"], losses_module.get
+            )
+        except ValueError:
+            pass  # unknown loss
         compile_kwargs["loss"] = unflatten_params(
             items=compile_kwargs["loss"],
             params=route_params(
                 init_params, destination="loss", pass_filter=set(), strict=False,
             ),
         )
-        compile_kwargs["metrics"] = _class_from_strings(
-            compile_kwargs["metrics"], metrics_module.get
-        )
+        try:
+            compile_kwargs["metrics"] = _class_from_strings(
+                compile_kwargs["metrics"], metrics_module.get
+            )
+        except ValueError:
+            pass  # unknown loss
         compile_kwargs["metrics"] = unflatten_params(
             items=compile_kwargs["metrics"],
             params=route_params(
@@ -422,8 +432,14 @@ class BaseWrapper(BaseEstimator):
 
     def _ensure_compiled_model(self) -> None:
         # compile model if user gave us an un-compiled model
-        if not (hasattr(self.model_, "loss") and hasattr(self.model_, "optimizer")):
-            self.model_.compile(**self._get_compile_kwargs())
+        if not (
+            getattr(self.model_, "loss", None)
+            and getattr(self.model_, "optimizer", None)
+        ):
+            self._compile_model(self._get_compile_kwargs())
+
+    def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
+        self.model_.compile(**compile_kwargs)
 
     def _fit_keras_model(
         self,
@@ -549,8 +565,14 @@ class BaseWrapper(BaseEstimator):
                 for x in [self.loss, self.model_.loss]
             ):  # filter out loss list/dicts/etc.
                 if default_val is not None:
-                    default_val = loss_name(default_val)
-                given = loss_name(self.loss)
+                    try:
+                        default_val = loss_name(default_val)
+                    except ValueError:
+                        pass
+                try:
+                    given = loss_name(self.loss)
+                except ValueError:
+                    return  # idk placeholder
                 got = loss_name(self.model_.loss)
                 if given != default_val and got != given:
                     raise ValueError(
@@ -1257,7 +1279,7 @@ class KerasClassifier(BaseWrapper):
         ] = "rmsprop",
         loss: Union[
             Union[str, tf.keras.losses.Loss, Type[tf.keras.losses.Loss], Callable], None
-        ] = None,
+        ] = "auto",
         metrics: Union[
             List[
                 Union[
@@ -1309,6 +1331,22 @@ class KerasClassifier(BaseWrapper):
             # check that this is not a multiclass problem missing categories
             target_type = type_of_target(self.classes_)
         return target_type
+
+    def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
+        if compile_kwargs["loss"] == "auto":
+            if len(self.model_.outputs) > 1:
+                raise ValueError(
+                    'Only single-output models are supported with `loss="auto"`'
+                )
+            if self.target_type_ == "binary":
+                compile_kwargs["loss"] = "binary_crossentropy"
+            else:
+                if self.model_.outputs[0].shape[1] == 1:
+                    raise ValueError(
+                        "Multi-class targets require the model to have >1 output unit"
+                    )
+                compile_kwargs["loss"] = "categorical_crossentropy"
+        self.model_.compile(**compile_kwargs)
 
     @staticmethod
     def scorer(y_true, y_pred, **kwargs) -> float:
@@ -1610,6 +1648,75 @@ class KerasRegressor(BaseWrapper):
         },
         **BaseWrapper._tags,
     }
+
+    def __init__(
+        self,
+        model: Union[None, Callable[..., tf.keras.Model], tf.keras.Model] = None,
+        *,
+        build_fn: Union[
+            None, Callable[..., tf.keras.Model], tf.keras.Model
+        ] = None,  # for backwards compatibility
+        warm_start: bool = False,
+        random_state: Union[int, np.random.RandomState, None] = None,
+        optimizer: Union[
+            str, tf.keras.optimizers.Optimizer, Type[tf.keras.optimizers.Optimizer]
+        ] = "rmsprop",
+        loss: Union[
+            Union[str, tf.keras.losses.Loss, Type[tf.keras.losses.Loss], Callable], None
+        ] = "auto",
+        metrics: Union[
+            List[
+                Union[
+                    str,
+                    tf.keras.metrics.Metric,
+                    Type[tf.keras.metrics.Metric],
+                    Callable,
+                ]
+            ],
+            None,
+        ] = None,
+        batch_size: Union[int, None] = None,
+        validation_batch_size: Union[int, None] = None,
+        verbose: int = 1,
+        callbacks: Union[
+            List[Union[tf.keras.callbacks.Callback, Type[tf.keras.callbacks.Callback]]],
+            None,
+        ] = None,
+        validation_split: float = 0.0,
+        shuffle: bool = True,
+        run_eagerly: bool = False,
+        epochs: int = 1,
+        **kwargs,
+    ):
+
+        # Parse hardcoded params
+        self.model = model
+        self.build_fn = build_fn
+        self.warm_start = warm_start
+        self.random_state = random_state
+        self.optimizer = optimizer
+        self.loss = loss
+        self.metrics = metrics
+        self.batch_size = batch_size
+        self.validation_batch_size = validation_batch_size
+        self.verbose = verbose
+        self.callbacks = callbacks
+        self.validation_split = validation_split
+        self.shuffle = shuffle
+        self.run_eagerly = run_eagerly
+        self.epochs = epochs
+
+        # Unpack kwargs
+        vars(self).update(**kwargs)
+
+        # Save names of kwargs into set
+        if kwargs:
+            self._user_params = set(kwargs)
+
+    def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
+        if compile_kwargs["loss"] == "auto":
+            compile_kwargs["loss"] = "mean_squared_error"
+        self.model_.compile(**compile_kwargs)
 
     @staticmethod
     def scorer(y_true, y_pred, **kwargs) -> float:
--- a/tests/mlp_models.py
+++ b/tests/mlp_models.py
@@ -26,11 +26,16 @@ def dynamic_classifier(
     for layer_size in hidden_layer_sizes:
         hidden = Dense(layer_size, activation="relu")(hidden)
 
+    if compile_kwargs["loss"] == "auto":
+        loss = None
+    else:
+        loss = compile_kwargs["loss"]
+
     if target_type_ == "binary":
-        compile_kwargs["loss"] = compile_kwargs["loss"] or "binary_crossentropy"
+        compile_kwargs["loss"] = loss or "binary_crossentropy"
         out = [Dense(1, activation="sigmoid")(hidden)]
     elif target_type_ == "multilabel-indicator":
-        compile_kwargs["loss"] = compile_kwargs["loss"] or "binary_crossentropy"
+        compile_kwargs["loss"] = loss or "binary_crossentropy"
         if isinstance(n_classes_, list):
             out = [
                 Dense(1, activation="sigmoid")(hidden)
@@ -39,13 +44,11 @@ def dynamic_classifier(
         else:
             out = Dense(n_classes_, activation="softmax")(hidden)
     elif target_type_ == "multiclass-multioutput":
-        compile_kwargs["loss"] = compile_kwargs["loss"] or "binary_crossentropy"
+        compile_kwargs["loss"] = loss or "binary_crossentropy"
         out = [Dense(n, activation="softmax")(hidden) for n in n_classes_]
     else:
         # multiclass
-        compile_kwargs["loss"] = (
-            compile_kwargs["loss"] or "sparse_categorical_crossentropy"
-        )
+        compile_kwargs["loss"] = loss or "sparse_categorical_crossentropy"
         out = [Dense(n_classes_, activation="softmax")(hidden)]
 
     model = Model(inp, out)
@@ -60,13 +63,13 @@ def dynamic_regressor(
     meta: Optional[Dict[str, Any]] = None,
     compile_kwargs: Optional[Dict[str, Any]] = None,
 ) -> Model:
-    """Creates a basic MLP regressor dynamically.
-    """
+    """Creates a basic MLP regressor dynamically."""
     # get parameters
     n_features_in_ = meta["n_features_in_"]
     n_outputs_ = meta["n_outputs_"]
 
-    compile_kwargs["loss"] = compile_kwargs["loss"] or "mse"
+    if compile_kwargs["loss"] == "auto":
+        compile_kwargs["loss"] = "mean_squared_error"
 
     inp = Input(shape=(n_features_in_,))
 
--- a/scikeras/utils/transformers.py
+++ b/scikeras/utils/transformers.py
@@ -154,7 +154,7 @@ class ClassifierLabelEncoder(BaseEstimator, TransformerMixin):
             "multiclass-multioutput": FunctionTransformer(),
             "multilabel-indicator": FunctionTransformer(),
         }
-        if is_categorical_crossentropy(self.loss):
+        if target_type == "multiclass" and is_categorical_crossentropy(self.loss):
             encoders["multiclass"] = make_pipeline(
                 TargetReshaper(),
                 OneHotEncoder(


scikeras/wrappers.py
```diff
@@ -345,27 +346,36 @@ def _get_compile_kwargs(self):
...
         )
-        compile_kwargs["loss"] = _class_from_strings(
-            compile_kwargs["loss"], losses_module.get
-        )
+        try:
+            compile_kwargs["loss"] = _class_from_strings(
+                compile_kwargs["loss"], losses_module.get
+            )
+        except ValueError:
+            pass  # unknown loss
```
adriangb, line 366: really only this one is needed, to pass through "auto". But this defense is probably a good idea on all of them.

scikeras/wrappers.py
```diff
@@ -549,8 +565,14 @@ def _check_model_compatibility(self, y: np.ndarray) -> None:
                 for x in [self.loss, self.model_.loss]
             ):  # filter out loss list/dicts/etc.
                 if default_val is not None:
-                    default_val = loss_name(default_val)
-                given = loss_name(self.loss)
+                    try:
+                        default_val = loss_name(default_val)
+                    except ValueError:
+                        pass
```
adriangb, line 571: Todo: figure out what to do here, or even refactor this check like in #208

scikeras/wrappers.py
```diff
@@ -1257,7 +1279,7 @@ def __init__(
         ] = "rmsprop",
         loss: Union[
             Union[str, tf.keras.losses.Loss, Type[tf.keras.losses.Loss], Callable], None
-        ] = None,
+        ] = "auto",
```
adriangb, line 1282: TODO: bring in the nice type-hint changes from #208 into a future PR (no need to put it in here)

scikeras/wrappers.py
```diff
@@ -1611,6 +1649,75 @@ class KerasRegressor(BaseWrapper):
...
+        self.shuffle = shuffle
+        self.run_eagerly = run_eagerly
+        self.epochs = epochs
+
+        # Unpack kwargs
+        vars(self).update(**kwargs)
+
+        # Save names of kwargs into set
+        if kwargs:
+            self._user_params = set(kwargs)
```
adriangb, line 1714: This is literally just to add a default value for loss.

scikeras/wrappers.py
```diff
@@ -1611,6 +1653,75 @@ class KerasRegressor(BaseWrapper):
...
+        model: Union[None, Callable[..., tf.keras.Model], tf.keras.Model] = None,
+        *,
+        build_fn: Union[
+            None, Callable[..., tf.keras.Model], tf.keras.Model
+        ] = None,  # for backwards compatibility
+        warm_start: bool = False,
+        random_state: Union[int, np.random.RandomState, None] = None,
+        optimizer: Union[
+            str, tf.keras.optimizers.Optimizer, Type[tf.keras.optimizers.Optimizer]
+        ] = "rmsprop",
```
stsievert, line 1667:
```suggestion
        ] = "adadelta",
```

scikeras/wrappers.py
stsievert: Why is `self._user_params` required? I'm not seeing any usage.

tests/mlp_models.py
```diff
@@ -26,11 +26,16 @@ def dynamic_classifier(
     for layer_size in hidden_layer_sizes:
         hidden = Dense(layer_size, activation="relu")(hidden)
 
+    if compile_kwargs["loss"] == "auto":
+        loss = None
+    else:
+        loss = compile_kwargs["loss"]
```
stsievert, line 32: Suggestion:

``` python
loss = None if compile_kwargs["loss"] == "auto" else compile["kwargs"]
```

scikeras/wrappers.py
```diff
@@ -422,8 +432,14 @@ def _build_keras_model(self):
 
     def _ensure_compiled_model(self) -> None:
         # compile model if user gave us an un-compiled model
-        if not (hasattr(self.model_, "loss") and hasattr(self.model_, "optimizer")):
-            self.model_.compile(**self._get_compile_kwargs())
+        if not (
+            getattr(self.model_, "loss", None)
+            and getattr(self.model_, "optimizer", None)
```
stsievert, line 437: Why is this change necessary?

scikeras/wrappers.py
stsievert: I think `loss_name` returning `None` says "the provided loss has no name/is not recognized."

scikeras/wrappers.py
stsievert: :+1: I added the types because it made `KerasRegressor.__init__` easier to implement.

I think `test_simple_usage.py` should also be brought in too.

tests/mlp_models.py
stsievert: Or should `None` be passed to `build_fn` if `loss="auto"` is specified?

adriangb: I see pros and cons.

Pros:
- Backwards compatibility.

Cons:
- `build_fn` cannot know if `None` was explicitly passed to the constructor or if `"auto"` was converted to None.

I'll add more if I can think of it, and try a quick implementation, it might work just fine...

scikeras/wrappers.py
adriangb: I'll try to put that change in. I still think this check needs to be refactored

scikeras/wrappers.py
adriangb: `test_simple_usage.py` is brought in.

I do like the types, but I think it can be done in a future PR to keep the intent here clear (even if it results in a larger diff because`__init__` is ugly).

scikeras/wrappers.py
adriangb: Why is this change necessary as part of this PR? This is just copied from `BaseWrapper.__init__`. If we change the default, we change the behavior.

scikeras/wrappers.py
adriangb: It is used by BaseWrapper. But I now realize I made a bonehead mistake and should have just called `super().__init__(...)`, which will in turn do this `self._user_params` logic. Not sure if that is what you were referring to or if you were referring to the existence of `self._user_params` altogether. If it's the latter, I think it's beyond the scope of this PR.

scikeras/wrappers.py
adriangb: Edit: I did not push `test_simple_usage.py`. I'll push it now. Sorry for the confusion.

scikeras/wrappers.py
stsievert: I'll submit another PR with type changes to `__init__`.

`"rmsprop"` is the default for [keras.model.compile][1], so I suppose it makes at least a little sense. But it's not at all popular.

[1]:https://keras.io/api/models/model_training_apis/#compile-method

adriangb: > I'll submit another PR with type changes to __init__.

Great, thank you! Please also feel free to just leave it there and I can get to it at some point, I do want those changes.

> it's not at all popular

I hear you. Indeed the only reason it's the default is because that is `Model.compile`'s default. A discussion to change this is welcome, but I think we'll need some more pros other than the fact that it's not that popular, and which one we should choose to replace it.

scikeras/wrappers.py
adriangb: Removed via https://github.com/adriangb/scikeras/pull/210/files#r587758575

scikeras/wrappers.py
adriangb: Good catch. This change isn't really necessary. I'll remove it.

scikeras/wrappers.py
adriangb: This worked out. Only small doc and test changes needed.

scikeras/wrappers.py
adriangb: Pushed.

scikeras/wrappers.py
adriangb: Replaced with a `super().__init__(...)` call.

tests/mlp_models.py
adriangb: > Suggestion:
> `loss = None if compile_kwargs["loss"] == "auto" else compile["kwargs"]`

Incorporated

adriangb (commented):
I incorporated most of the feedback. I think the main outstanding issue is https://github.com/adriangb/scikeras/pull/210#discussion_r587760368

scikeras/wrappers.py
adriangb: Tracking type-hints in https://github.com/adriangb/scikeras/issues/211

poetry.lock
```diff
@@ -309,21 +309,22 @@ six = ">=1.9.0"
 
 [package.extras]
 aiohttp = ["aiohttp (>=3.6.2,<4.0.0dev)"]
+pyopenssl = ["pyopenssl (>=20.0.0)"]
```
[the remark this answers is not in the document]
stsievert: Should this file even be on GitHub?

scikeras/wrappers.py
stsievert: I think the consistency with Keras trumps any concerns I have.

scikeras/wrappers.py
```diff
@@ -1310,6 +1322,26 @@ def _type_of_target(self, y: np.ndarray) -> str:
...
+                compile_kwargs["loss"] = "binary_crossentropy"
+            elif self.target_type_ == "multiclass":
+                if self.model_.outputs[0].shape[1] == 1:
+                    raise ValueError(
+                        "Multi-class targets require the model to have >1 output unit"
+                    )
+                compile_kwargs["loss"] = "sparse_categorical_crossentropy"
+            else:
+                raise ValueError(
+                    f'`loss="auto"` is not supported for tasks of type {self.target_type_}'
```
stsievert, line 1341: When is this error triggered? Is there a test for it?

tests/test_simple_usage.py
```diff
@@ -0,0 +1,135 @@
...
+    assert est.current_epoch == 1
+
+
+@pytest.mark.parametrize(
+    "use_case,supported",
+    [
+        ("binary_classification", True),
+        ("binary_classification_w_one_class", True),
+        ("classification_w_1d_targets", True),
+        ("classification_w_onehot_targets", False),
```
stsievert, line 78: Isn't classification with one-hot targets a really important use case that should be supported?

tests/test_simple_usage.py
```diff
@@ -0,0 +1,135 @@
...
+    elif loss == "categorical_crossentropy":
+        # SciKeras does not auto one-hot encode unless
+        # loss="categorical_crossentropy" is explictily passed to the constructor
+        y = np.random.randint(0, N_CLASSES, size=(n_eg, 1))
+        y = OneHotEncoder(sparse=False).fit_transform(y)
+    else:
+        y = np.random.randint(0, N_CLASSES, size=(n_eg,))
+    est = KerasClassifier(
+        shallow_net,
+        model__compile=True,
```
stsievert, line 62:
``` python
@pytest.mark.parametrize("compile", [True, False])
```

?

tests/test_simple_usage.py
```diff
@@ -0,0 +1,135 @@
...
+    est.partial_fit(X, y)
+    assert loss_name(est.model_.loss) == "mean_squared_error"
+
+
+@pytest.mark.parametrize(
+    "use_case,loss",
+    [
+        ("binary_classification", "binary_crossentropy"),
+        ("binary_classification_w_one_class", "binary_crossentropy"),
+        ("classification_w_1d_targets", "sparse_categorical_crossentropy"),
```
stsievert, line 121: Where's classification with one-hot targets?

tests/test_simple_usage.py
stsievert: It might be worth collapsing these tests. They're all very similar, and I'm having a hard time telling the difference.

poetry.lock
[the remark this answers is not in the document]
adriangb: This is from me typing stuff in the wrong terminal. I'll revert this.

Regarding the file itself, some libraries keep them around, some don't. Lately I've come around to the camp of not keeping it around, so I'll probably remove it down the road.

poetry.lock
```diff
@@ -295,7 +295,7 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "google-auth"
```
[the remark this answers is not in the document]
adriangb: You can't specify dev extras. But I agree it may be easier to just make them dev deps. But let's table that for another PR, I'm going to revert this change here anyways.

poetry.lock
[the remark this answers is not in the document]
stsievert: What use does it provide?

scikeras/wrappers.py
adriangb: If the user passes multilabel indicator, multiclass multioutput, etc.

No test yet, I'll put one in tomorrow

tests/test_simple_usage.py
adriangb: It will be supported, but only if the user explicitly passes the loss function. That is tested elsewhere.

adriangb:
For deployable things, like a containerized servicie, it makes builds
deterministic and faster.

For libraries, the use is more questionable.

I think the Poetry authors can probably offer better responses than I can:
https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control

I think they've changed their wording: I think it used to say libraries MAY
not commit it, but now says SHOULD not commit it.

In any case, we should remove it.

poetry.lock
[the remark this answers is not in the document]
adriangb: For deployable things like a containerized servicie, it makes builds
deterministic and faster.

For libraries, the use is more questionable.

The Poetry authors can probably offer better responses than I can:
https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control

I think they've changed their wording: I think it used to say libraries MAY
not commit it, but now says SHOULD not commit it.

tests/test_simple_usage.py
adriangb: This is not really a change, since this was not supported by `loss=None` either.

poetry.lock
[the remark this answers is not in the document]
stsievert: Filed #212.

scikeras/wrappers.py
stsievert: Why can't that use case be supported?

adriangb: Multiclass-multioutput can't be supported because we have no way of knowing what loss users want for each output, or even how many outputs there are.

multilabel-indicator can't be differentiated from one-hot encoded targets, you'd need to introspect the model architecture.

These are the same reasons why LabelEncoder does not support these use cases, and the same reason why one-hot encoded targets aren't supported if the user doesn't pass an explicit single loss function and that loss function is cce (which is the only way we can know to one-hot encode the targets without introspecting the model).

We could set up some type of sentinel value from LabelEncoder to handle this, but this produces tight coupling, which I do not like

scikeras/wrappers.py
```diff
@@ -1310,6 +1322,26 @@ def _type_of_target(self, y: np.ndarray) -> str:
...
+            if len(self.model_.outputs) > 1:
+                raise ValueError(
+                    'Only single-output models are supported with `loss="auto"`'
+                )
+            if self.target_type_ == "binary":
+                compile_kwargs["loss"] = "binary_crossentropy"
+            elif self.target_type_ == "multiclass":
+                if self.model_.outputs[0].shape[1] == 1:
+                    raise ValueError(
+                        "Multi-class targets require the model to have >1 output unit"
```
stsievert, line 1336:
```suggestion
                        f"Multi-class targets require the model to have >1 output unit instead of {self.model_.outputs[0].shape} units"
```

scikeras/wrappers.py
stsievert: I'm most concerned with one-hot encoded targets, or multiclass-indicator. That's an extremely common usage.

Why can't the loss default to `categorical_crossentropy`, and change that default value in special cases?

stsievert: For ease of documentation, I'd almost rather the loss defaulted to `categorical_crossentropy`.

adriangb: If we do that, we can't support non one-hot encoded multiclass targets since at the point they'd have to be one-hot encoded (LabelClassifier) it isn't known if the user is going to to give us a compiled or uncompiled model.

adriangb: We could introduce a `"multiclass-indicator"` target type but it is a bit dangerous because it can be confused with a multilabel indicator where no samples have multiple classes, and they are very different problems.

adriangb: We can certainly improve the error message to tell users exactly what's going on and how to resolve it. Would that help alleviate the concern?

scikeras/wrappers.py
adriangb: Reverted

tests/test_simple_usage.py
adriangb: I attempted to collapse these tests. Let me know if it is clearer now.

scikeras/wrappers.py
adriangb: There is a test for this error now.

Note that the initialization flow won't even make it this far unless the user is already subclassing and overriding the default target_encoder, since the default implementation (`LabelEncoder`) will raise its own `NotImplementedError` for these same cases. The only case that will not raise an error until this point is one-hot-encoded targets.

For one-hot-encoded targets, the error will look as follows:

```
`loss="auto"` is not supported for tasks of type multilabel-indicator.
Instead, you must explicitly pass a loss function, for example:
  clf = KerasClassifier(..., loss="categorical_crossentropy")
```

tests/test_api.py
```diff
@@ -327,7 +327,7 @@ def test_basic(self, config):
             keras_model = build_fn(
                 meta=meta,
                 hidden_layer_sizes=(100,),
-                compile_kwargs={"optimizer": "adam", "loss": None, "metrics": None,},
+                compile_kwargs={"optimizer": "adam", "loss": "auto", "metrics": None,},
```
adriangb, line 330: These tests are directly calling the model building function which is debatably not a great way to set up tests. This just changes the inputs to the model building function to reflect what it would now be getting if users don't specify a loss function.

adriangb:
@stsievert do you think we should move forward with this PR?

scikeras/wrappers.py
```diff
@@ -1611,6 +1661,73 @@ class KerasRegressor(BaseWrapper):
...
+            run_eagerly=run_eagerly,
+            epochs=epochs,
+            **kwargs,
+        )
+
+    def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
+        if compile_kwargs["loss"] == "auto":
+            if len(self.model_.outputs) > 1:
+                raise ValueError(
+                    'Only single-output models are supported with `loss="auto"`'
```
stsievert, line 1726: Does this agree with the documentation?

> Regression always defaults to mean squared error. For multi-output models, Keras will use the sum of each output's loss.
> —https://github.com/adriangb/scikeras/pull/210/files#diff-a330a0112e60c2872ba1c9bd84f85a963f9edc44a273d883fed5b59c5e8b4a98R167

docs/source/advanced.rst
```diff
@@ -128,6 +128,43 @@ offer an easy way to compile and tune compilation parameters. Examples:
...
+excellent article `How to Choose Loss Functions When Training Deep Learning Neural Networks`_
+as well as `Keras Losses docs`_.
+
+Default losses are selected as follows:
+
+Classification
+..............
+
++-----------+-----------+----------+---------------------------------+
+| # outputs | # classes | encoding | loss                            |
```
stsievert, line 149: I'm confused by this table. Let's say I have two classes, one "output," and I don't know my "encoding" (I'm not sure a naive user would know what that means). What loss is chosen?

Maybe it'd be simpler to say "KerasClassifier has `loss="sparse_categorical_crossentropy"` by default. It works for one dimensional labels like `est.fit(X, [2, 3, 4, 5])`). If you have binary labels like `y=[-1, 1, -1, -1]`, specify `binary_crossentropy`. If you have one-hot encoded labels, use LOSS."

docs/source/advanced.rst
```diff
@@ -128,6 +128,43 @@ offer an easy way to compile and tune compilation parameters. Examples:
...
+that matches the type of target (see :py:func:`sklearn.utils.multiclass.type_of_target`).
+
+For guidance selecting losses in Keras, please see Jason Brownlee's
+excellent article `How to Choose Loss Functions When Training Deep Learning Neural Networks`_
+as well as `Keras Losses docs`_.
+
+Default losses are selected as follows:
+
+Classification
+..............
```
stsievert, line 146: I think this section could use some use examples, and clarification of what "output" and "encoding" mean.

docs/source/advanced.rst
adriangb: How about:

> KerasClassifier will automatically determine an appropriate loss function for binary (`[0, 1, 0]`/`["car", "bike", "car"]`) or multiclass (`[1, 2, 3, 4]`/`["person", "car", "pear", "tree"]`) one-dimensional targets. For other types of target, you must explicitly specify the loss. If your target is one-hot encoded, you probably want to use `"categorical_crossentropy"`.

docs/source/advanced.rst
adriangb: > In this context, `outputs` refers to the number of things you are predicting (for example, you could predict just `color`, in which case you have 1 output, or you might predict `color` and `is_tshirt`, in which case you have 2 outputs). Encoding refers to the representation of the target data. Generally, you will see data encoded as labels (`[1, 2, 3]` or `["red", "green", "blue"]`) or one-hot encoded. See [one-hot on Wikipedia](https://en.wikipedia.org/wiki/One-hot) for more details.

scikeras/wrappers.py
adriangb: Good catch. I think the documentation is wrong. I think we shouldn't support multi-output at all, there's nothing to say that the sum is the right way to aggregate them (although that is what Keras does by default...).

stsievert: I think SciKeras should mirror Keras as closely as possible.

docs/source/advanced.rst
stsievert: I'm only for using `loss="auto"` if there are simple and easy-to-follow rules. Setting a fixed value of `loss="sparse_categorical_crossentropy"` is a really simple rule.

I almost prefer this documentation:

> KerasClassifier has `loss="sparse_categorical_crossentropy"` by default. This assumes that the model has C outputs neurons to classify C classes. It's intended to be used like this:
>
> ``` python
> def build_model():
>     ...
>     model.add(output_layer_C_neurons)
>     return model
> 
> est = KerasClassifier(model=build_model)
> est.fit(X, [0, 1, 2, 0, 1, 2])
> ```
> 
> If you have one-hot encoded targets, manually 
>
> ``` python
> from sklearn.datasets import OneHotEncoder
> est = KerasClassifier(model=build_model, loss="categorical_crossentropy")
> y = OneHotEncoder([0, 1, 2, 0, 1, 2])
> y = [[1, 0, 0], [0, 1, 0], [0, 0, 1], [1, 0, 0], [0, 1, 0], [0, 0, 1]]  # or this
> est.fit(X, y)
> ```

adriangb: I think we tried this before. I don't remember the conclusion of those discussions (although I can dig it up), but off the top of my head I think the biggest issue is that new users will copy an example model from a tutorial, many of which do binary classification using a single neuron, or other incompatible architectures.

Do you think we can just introspect the model and check if the number of neurons matches the number of classes (and that it is a single-output problem) and raise an error (or maybe a warning) to rescue users from facing whatever cryptic error TF would throw?

stsievert: > I don't remember the conclusion of those discussions (although I can dig it up

I recall introspecting the model to see what loss value should be used, but trying to abstract too much away from the user (and plus it got too complicated).

I think the new loss for `KerasClassifier` is better: it's very simple and recommend changes if common mistakes are made (eventually; see below).

> introspect the model and ... raise an error (or maybe a warning)

Yeah, I had the same idea. If I were developing this library, I think I'd have `loss="sparse_categorical_crossentropy"` with clear documentation ("have model return one neuron for each output, likely with softmax activation"). I would catch these use cases:

* 1 output neuron and `loss != "binary_crossentropy"`.
* target one-hot encoded (and tell to set loss="categorical_crossentropy"`).

I think both of these should be exceptions. If so, I'd make it clear how to clear how to adapt to BaseWrapper.

> copy an example model from a tutorial, many of which do binary classification using a single neuron

I think a clear documentation note would resolve this, especially with good error catching.

<details>
<summary>keras.io examples</summary>

* One neuron: [3D CT scans][3], [text classification][7]
* Dynamic choice: [image classification][1], 
* ≥ 2 neurons: [simple mnist][2], [EfficientNet][4], [ViT][5], [Perceiver][6], [text w/ switch][8], [text w/ transformer][9a]

</summary>

[1]:https://keras.io/examples/vision/image_classification_from_scratch/#build-a-model
[2]:https://keras.io/examples/vision/mnist_convnet/#build-the-model
[3]:https://keras.io/examples/vision/3D_image_classification/#define-a-3d-convolutional-neural-network
[4]:https://keras.io/examples/vision/image_classification_efficientnet_fine_tuning/#training-a-model-from-scratch
[5]:https://keras.io/examples/vision/image_classification_with_vision_transformer/#build-the-vit-model
[6]:https://keras.io/examples/vision/perceiver_image_classification/#prepare-the-data
[7]:https://keras.io/examples/nlp/text_classification_from_scratch/#build-a-model
[8]:https://keras.io/examples/nlp/text_classification_with_switch_transformer/#implement-the-classifier
[9a]:https://keras.io/examples/nlp/text_classification_with_transformer/#create-classifier-model-using-transformer-layer

scikeras/wrappers.py
stsievert: (that is, I think the total loss should be the sum of outputs; that's what Keras does).

docs/source/advanced.rst
adriangb: I think https://github.com/adriangb/scikeras/pull/210#discussion_r627916340 is at at least worth exploring (again).

I'll open a new PR to test out https://github.com/adriangb/scikeras/pull/210#discussion_r627916340 and to avoid loosing the changes here into the git history, and also because the changes are going to be pretty unrelated.

Thank you for following up on this PR 😄

PR #218, opened 2021-03-08
MAINT: better error message about one-hot encoded targets w/ loss="auto"

stsievert:
It adds support for one-hot encoded targets when `loss="auto"`. If the classifier has the attributed `n_classes_` (which `KerasClassifier.target_encoder_` provides by default), then `loss="categorical_crossentropy"` is provided.

**Reference issues/PRs**
#208 I guess.


adriangb:
This is great, good idea! A lot of my concerns are mitigated by the fact that:
 - It only works with pre-onhotted targets
 - If `n_classes_` is absent, users will still get a `NotImplementedError` instead of possibly confusing results.

scikeras/wrappers.py
```diff
@@ -1340,6 +1340,15 @@ def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
                         "Multi-class targets require the model to have >1 output units."
                     )
                 compile_kwargs["loss"] = "sparse_categorical_crossentropy"
+            elif hasattr(self, "n_classes_"):
+                n_out = self.model_.outputs[0].shape[1]
+                if n_out != self.n_classes_:
+                    raise ValueError(
+                        "loss='categorical_crossentropy' is expecting the model "
+                        f"to have {self.n_classes_} output neurons, one for each "
+                        "class. However, only {n_out} output neurons were found"
```
stsievert, line 1349:
```suggestion
                        f"class. However, only {n_out} output neurons were found"
```

stsievert:
This PR still needs a test to catch when the user mis-specifies the number of output neurons.

adriangb:
> This PR still needs a test to catch when the user mis-specifies the number of output neurons.

That is, to check for [this](https://github.com/adriangb/scikeras/blob/3be3c26505b646eaff54d789f85d95037e39ee2d/scikeras/wrappers.py#L1346-L1350) `ValueError`?

scikeras/wrappers.py
```diff
@@ -1340,6 +1340,15 @@ def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
                         "Multi-class targets require the model to have >1 output units."
                     )
                 compile_kwargs["loss"] = "sparse_categorical_crossentropy"
+            elif hasattr(self, "n_classes_"):
```
adriangb, line 1343:
```suggestion
            elif self.target_type_ == "multilabel-indicator" and hasattr(self, "n_classes_"):
```

adriangb:
Would you be open to, instead of raising the specific ValueError, allow the more general `this task is not supported` error below to be raised for `unsupported task of type 'multilabel-indicator', since we basically aren't sure that this is a one-hot problem?

stsievert:
Are you talking adding the message "loss='auto' is not supported for tasks of type 'multilabel-indicator'" to [this][1] ValueError?

[1]:https://github.com/adriangb/scikeras/pull/218/commits/fc9e7ee98a8c4eec95c6f68b3629e31ce3874131#diff-c0fb7ade08b2fcb7b16bd44c615a644896f3a6c505a4cf3fcb0545e4d6b4c3dfR1348-R1352

adriangb:
I'm talking about a flow like this:

```python

if loss == "auto":
  activation = self.model_.layers[-1].activation.__name__
  ...
  elif (
      self.target_type_ == "multilabel-indicator"
      and
      hasattr(self, "n_classes_")
      and
      self.model_.outputs[0].shape[1] == self.n_classes_
      and
      activation in ("linear", "softmax")
  ):
      use_logits = activation == "linear"
      compile_kwargs["loss"] = CategoricalCrossentropy(from_logits=use_logits)
  else:
      raise NotImplementedError(
          f'`loss="auto"` is not supported for tasks of type {self.target_type_}.'
          " Instead, you must explicitly pass a loss function, for example:"
          '\n   clf = KerasClassifier(..., loss="categorical_crossentropy")'
      )
```

Basically, we check if the combination of the target-type and model outputs indicate with high certainty that this is a run of the mill multiclass problem with one-hot encoded labels and if not, we tell the user that we don't know how to handle this type of target (which would then be classified as `multilabel-indicator').

stsievert:
I think checks for the `n_classes_` attribute is enough, especially if the number of output nodes is the number of classes. I also think that's simpler to document and for the users to understand. I think as a user, I'd rather see a documentation page explaining what activation functions to use.

[1]:https://keras.io/api/layers/activations/

adriangb:
This use case:

```python
def get_model(compile):
    model = keras.Sequential(
        [
            keras.Input((num_classes,)),
            keras.layers.Dense(num_classes, activation="relu"),
            keras.layers.Dense(num_classes, activation="sigmoid"),
        ]
    )
    if compile:
        model.compile(loss="binary_crossentropy")
    return model
```

This is used when each class is independent and there can be more than 1 class per label, for example to characterize sentiment in text or movie genres.

<details>
<summary>Full example code on fc9e7ee</summary>

```python
import numpy as np
from tensorflow import keras

num_classes = 3
n_samples = 10000

y = np.random.randint(low=0, high=2, size=(n_samples, num_classes))
X = y

def get_model(compile):
    model = keras.Sequential(
        [
            keras.Input((num_classes,)),
            keras.layers.Dense(num_classes, activation="relu"),
            keras.layers.Dense(num_classes, activation="sigmoid"),
        ]
    )
    if compile:
        model.compile(loss="binary_crossentropy")
    return model

m1 = get_model(True)
m1.fit(X, y, epochs=10)

m2 = KerasClassifier(get_model, model__compile=False, epochs=10).fit(X, y)
assert m2.model_.loss == "categorical_crossentropy"
```

</details>

Even worse than an error, there is no error, SciKeras silently chooses `"categorical_crossentropy"`, which is wrong for this problem, resulting in no / bad learning with no errors and no way to really tell what happened based on the output (it looks fine in terms of shape, dtype and being binary values, it's just a bunch of garbage).

I'm not sure we should support this use case via `loss="auto"` (although MLPClassifier does and our LabelEncoder does as well, so perhaps we should support it here), but I am sure we should not assume we can use `"categorical_crossentropy"` for this problem.

stsievert:
Why not ensure that one-hot encoded labels are passed? The target encoder should probably do that verification.

adriangb:
What do you mean by ensure that one-hot labels are passed? What if LabelEncoder gets non one-hot labels? If `loss="auto"`, how is it supposed to do in this situation? Currently, it does nothing, which I think is the right behavior.

stsievert:
Maybe it'd be better to raise an error and provide a some information on why a user might not want to set `loss="categorical_crossentropy"`.

adriangb:
> Maybe it'd be better to raise an error and provide a some information on why a user might not want to set `loss="categorical_crossentropy"`.

Where would this error be raised?

Am I understanding correctly that you feel that this extra bit of introspection into the model is too much, and that instead we should not support any of the use cases in https://github.com/adriangb/scikeras/pull/218#issuecomment-792828867?

stsievert:
I think a warning should be issued if SciKeras thinks `loss="categorical_crossentropy"` is appropriate. I think an error should be raised if a loss is not chosen. I think this would be an appropriate error message:

> SciKeras could not automatically decide on the appropriate loss. To clear this error, set the loss to be any valid argument Keras accepts for losses (details at https://keras.io/api/losses).
>
> The target type is "multilabel-indicator," there are 4 classes and the model has 4 output neurons. If there is only one class per example, loss="categorical_crossentropy" might be appropriate. If there are multiple classes per example, loss="binary_crossentropy" might be appropriate. For more detail, see XXX.

I think the last paragraph should be added dynamically.

> Am I understanding correctly that you feel that this extra bit of introspection into the model is too much, and that instead we should not support any of the use cases in #218 (comment)?

Yes. I think the user knows their model/dataset better than SciKeras, and they're also a lot smarter than SciKeras. I think SciKeras should support them (provide links/good information) in guiding their choice.

adriangb:
I agree on the error message if no loss is chosen. We can discuss the specifics of what that error message should contain separately. I'd first like to settle on what _is_ supported, that is, what you mean by `SciKeras thinks loss="categorical_crossentropy" is appropriate`.  How does Scikeras know that `"categorical_crossentropy"` is appropriate? Some pseudocode might help me understand, I'm sorry if I'm being slow to pick up.

stsievert:
``` python
def _compile(...):
    warning = ""
    loss = None
    if condition:
        loss = "binary_crossentropy"
    elif target_type == "multiclass-indicator" and hasattr(self, "n_classes_"):
        n_out = ...
        if n_out == self.n_classes_:
            warning = (
                "The target type is 'multilabel-indicator,' there are 4 classes and the model has 4 output neurons. "
                "If there is only one class per example, loss='categorical_crossentropy' might be appropriate. "
                "If there are multiple classes per example, loss='binary_crossentropy' might be appropriate. "
                "For more detail, see XXX."
            )
    if loss is None:
        err_msg = "loss='auto' failed to find a suitable loss. To clear this message, set loss to be a Keras loss ..."
        if len(warning):
            warning = f"\n\n{warning}" 
        raise ValueError("{err_msg}{warning}")
    else:
        compile_kwargs["loss"] = loss
```

adriangb:
So the `elif target_type == "multiclass-indicator" and hasattr(self, "n_classes_")` does not set `loss`? This means that one-hot encoded targets would not be supported (although they will get a helpful error). I am okay with that; I think we can always add support later based on user feedback.

stsievert:
> So the elif target_type == "multilabel-indicator" and hasattr(self, "n_classes_") does not set loss? This means that one-hot encoded targets would not be supported (although they will get a helpful error). I am okay with that; I think we can always add support later based on user feedback.

Yeah, that's too much I think. It requires too introspection, or might produce unexpected results for the user. Users are smarter than software, so let's let them decide. I see why now you were hesitant to support it.

This PR should be refactored. I could use some help there.

adriangb:
> Users are smarter than software, so let's let them decide

👍 

> This PR should be refactored

If you don't mind me pushing to your fork, I can make the edits to have this reflect the intention in https://github.com/adriangb/scikeras/pull/218#issuecomment-792888899. How does that sound?

stsievert:
> If you don't mind me pushing to your fork, I can make the edits to have this reflect the intention in #218 (comment). How does that sound?

Please. Thanks for asking. I have the box "allow edits by maintainers" checked for a reason.

adriangb:
@stsievert the basics are here, do you want to look at this here, or as part of #210 ?

scikeras/wrappers.py
```diff
@@ -1328,24 +1327,37 @@ def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
...
-                compile_kwargs["loss"] = "sparse_categorical_crossentropy"
-            else:
-                raise NotImplementedError(
+                loss = "sparse_categorical_crossentropy"
+            elif self.target_type_ == "multilabel-indicator":
+                # one-hot encoded multiclass problem OR multilabel-indicator problem
+                hint = (
+                    "For this type of problem, the following may help:"
+                    '\n - If there is only one class per example, loss="categorical_crossentropy" might be appropriate.'
+                    '\n - If there are multiple classes per example, loss="binary_crossentropy" might be appropriate.'
```
stsievert, line 1349: I think SciKeras should have a documentation page that explains this more, and points to other resources.

scikeras/wrappers.py
```diff
@@ -1328,24 +1327,37 @@ def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
...
+                )
+            if loss is None:
+                msg = (
                     f'`loss="auto"` is not supported for tasks of type {self.target_type_}.'
-                    " Instead, you must explicitly pass a loss function, for example:"
+                    "\nInstead, you must explicitly pass a loss function, for example:"
```
stsievert, line 1354: Style nit:

```suggestion
                    " Instead, you must explicitly pass a loss function, for example:"
```

scikeras/wrappers.py
```diff
@@ -1328,24 +1327,37 @@ def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
...
                     '\n   clf = KerasClassifier(..., loss="categorical_crossentropy")'
                 )
+                if hint:
+                    msg += f"\n{hint}"
```
stsievert, line 1358: Style nit:

```suggestion
                    msg += f"\n\n{hint}"
```

scikeras/wrappers.py
```diff
@@ -1328,24 +1327,37 @@ def _compile_model(self, compile_kwargs: Dict[str, Any]) -> None:
...
                     raise ValueError(
                         "Multi-class targets require the model to have >1 output units."
                     )
-                compile_kwargs["loss"] = "sparse_categorical_crossentropy"
-            else:
-                raise NotImplementedError(
+                loss = "sparse_categorical_crossentropy"
+            elif self.target_type_ == "multilabel-indicator":
+                # one-hot encoded multiclass problem OR multilabel-indicator problem
+                hint = (
```
stsievert, line 1346: Maybe this hint should be a warning. Opinions?

tests/test_loss_auto.py
```diff
@@ -79,40 +79,53 @@ def target_encoder(self):
...
     by `loss="auto"`.
     """
+    extra = ""
     if use_case == "multiclass-multioutput":
         y1 = np.random.randint(0, 1, size=len(X))
         y2 = np.random.randint(0, 2, size=len(X))
         y = np.column_stack([y1, y2])
     elif use_case == "multilabel-indicator":
         y1 = np.random.randint(0, 1, size=len(X))
         y = np.column_stack([y1, y1])
```
stsievert, line 101:
```suggestion
        y = np.column_stack([y1, y1])
        assert np.unique(y.sum(axis=1)) == [0, 1, 2]
```

Is that appropriate?

tests/test_loss_auto.py
```diff
@@ -79,40 +79,53 @@ def target_encoder(self):
...
+    elif use_case == "classification_w_onehot_targets":
+        y = np.random.choice(N_CLASSES, size=len(X)).astype(int)
+        y = OneHotEncoder(sparse=False).fit_transform(y.reshape(-1, 1))
+        extra = 'loss="categorical_crossentropy" might be appropriate'
+    est = wrapper_cls(shallow_net, model__compile=False)
+    match = '`loss="auto"` is not supported for tasks of type'
+    if extra:
+        match += f"(.|\n)+{extra}"
+    with pytest.raises(NotImplementedError, match=match):
         est.initialize(X, y)
```
stsievert, line 112: I think if the error message suggests a loss value, that suggestion should be tested.

scikeras/wrappers.py
stsievert: I think there are two ways to clear this error:

1. Compile the model yourself (even if loss="auto")
2. Provide an explicit loss yourself (not loss="auto")

Is that accurate?

tests/test_loss_auto.py
```diff
@@ -214,7 +219,7 @@ def test_multiclass_single_output_unit():
 def test_binary_multiple_output_units():
```
stsievert, line 219: Can't binary classification work with loss="categorical_crossentropy"? If so, shouldn't that information be conveyed in the error message?

tests/test_loss_auto.py
```diff
@@ -202,7 +207,7 @@ def test_multi_output_support(user_compiled, est_cls):
 def test_multiclass_single_output_unit():
     """Test that multiclass targets requires > 1 output units.
     """
-    est = KerasClassifier(model=shallow_net, model__single_output=True)
+    est = KerasClassifier(model=shallow_net, model__outputs=1)
     y = np.random.choice(N_CLASSES, size=(len(X), 1)).astype(int)
     with pytest.raises(
         ValueError,
```
stsievert, line 213: Suggested change:

``` diff
-         match="Multi-class targets require the model to have >1 output units",
+        match="Targets with {n_classes} require the model to have {n_classes} output units, not {n_outputs} outputs",
```

scikeras/wrappers.py
adriangb: I agree, however I'd like to table this until we merge this PR into #210

scikeras/wrappers.py
adriangb: Yes, we can add a mention of the first, but not more than that, users should know what to do.

tests/test_loss_auto.py
adriangb: Good catch. I think the following is better:

```suggestion
        y = np.column_stack([y1, y1])
        y[0, :] = 1
```

tests/test_loss_auto.py
adriangb: I'm not sure how much this helps. There can only be 1 or >1 output units. By saying you need >1, it is clear that you must have 1, but if needed we can print `require the model to have >1 output units, but this model has only 1 output unit`. But I don't think it makes sense to make this dynamic, which would also require pulling in the `n_classes_` attribute, handling when it is not there, etc.

scikeras/wrappers.py
adriangb: If this hint exists, then an error is about to be raised (and this text is part of it). It wouldn't make sense to emit a warning right before raising an error, the warning would get lost. Unless I am missing something?

scikeras/wrappers.py
stsievert: Yes, we should definitely mention it in the error message. Maybe there should be a note about how it's only suitable for advanced usage.

tests/test_loss_auto.py
stsievert: This should be clearly mentioned on the documentation. I think that's appropriate for this PR.

scikeras/wrappers.py
adriangb: I added links to:
- https://machinelearningmastery.com/how-to-choose-loss-functions-when-training-deep-learning-neural-networks/
- https://www.tensorflow.org/api_docs/python/tf/keras/losses

In the SciKeras docs.

docs/source/quickstart.rst
```diff
@@ -40,14 +40,17 @@ it on a toy classification dataset using SciKeras
...
-        loss="sparse_categorical_crossentropy",
         hidden_layer_dim=100,
     )
 
     clf.fit(X, y)
     y_proba = clf.predict_proba(X)
 
 
+Note that SciKeras even chooses a loss function and compiles your model
+(see the :ref:`Advanced Usage` section for more details)!
```
stsievert, line 51: Suggested change: "if you want to add Keras's [loss name], specify `loss="loss_name"`".

adriangb: Thanks. I copied over the relevant section from the "Advanced" section instead:


```
To override the default loss, simply specify a loss function:

.. code-block:: diff

    -KerasClassifier(model=model_build_fn)
    +KerasClassifier(model=model_build_fn, loss="categorical_crossentropy")
```

Do you think that works?

tests/test_loss_auto.py
adriangb: Here is the new section in the docs: https://www.adriangb.com/scikeras/refs/pull/218/merge/advanced.html#loss-selection

docs/source/quickstart.rst
stsievert: Isn't `categorical_crossentropy` the default, or one of the defaults? Maybe use `loss="kl_divergence"` instead?

adriangb: It's one of the unsupported cases for `loss="auto"`. But at the same time, it is relatively common in Keras. So I chose it because it's unsupported but also common.

adriangb: I added some detail.

adriangb:
@stsievert I think we're ready to merge this back into #208 and continue there if you are okay with that

docs/source/advanced.rst
```diff
@@ -128,6 +128,40 @@ offer an easy way to compile and tune compilation parameters. Examples:
...
+| # outputs | # classes | encoding | loss                            |
++===========+===========+==========+=================================+
+| 1         | <= 2      | any      | binary crossentropy             |
++-----------+-----------+----------+---------------------------------+
+| 1         | >=2       | labels   | sparse categorical crossentropy |
++-----------+-----------+----------+---------------------------------+
+| 1         | >=2       | one-hot  | unsupported                     |
++-----------+-----------+----------+---------------------------------+
+| > 1       |    --     |    --    | unsupported                     |
++-----------+-----------+----------+---------------------------------+
```
stsievert, line 158: I'd also add a note here saying "if you have more than one output, `loss="categorical_crossentropy"` is likely desired" with maybe some links to an MNIST example or something.

docs/source/advanced.rst
```diff
@@ -1,6 +1,6 @@
-===================================
-Advanced Usage of SciKeras Wrappers
-===================================
+==============
+Advanced Usage
```
stsievert, line 2: Why is the section on loss selection in "advanced usage"? I suspect the average Keras user knows about loss/optimization.

adriangb: That is what it was historically named... I only shortened the name here in an attempt to get links working (and because the `of SciKeras Wrappers` part is arguably redundant). I'm happy to change it, in another PR.

docs/source/advanced.rst
adriangb: Maybe:
> If your targets are one-hot encoded, `loss="categorical_crossentropy"` is likely desired.
?

stsievert: :+1:

Issue #221, opened 2021-03-22
remove tensorflow dependency


fcakyon:
Due to the [tensorflow dependency](https://github.com/adriangb/scikeras/blob/d941d96040e1dda7bf2ae4b9fd861266d90f051b/pyproject.toml#L35), this package is not usable with [tensorflow-cpu](https://pypi.org/project/tensorflow-cpu/) package.

Can you please make tensorflow dependency optional so that I can use it with tensorflow-cpu?

adriangb:
Hmm, there's basically no description on the PyPi package (other than some generic `what is TensorFlow` stuff). What is the `tensorflow-cpu` package, and how does it differ from the plain `tensorflow` package? Or can you link me to any relevant docs? Thanks!

fcakyon:
@adriangb It is also maintained by Google and has the same api with tensorflow package. It doesnt include the compiled cuda binaries so that it takes less space and easier to deploy on machines without gpu.

fcakyon:
It is especially useful when deploying an ML app to Heroku: https://stackoverflow.com/a/62356779

adriangb:
Interesting, I sure with the TensorFlow team had better descriptions. Also if they packaged it as an extra like `pip install tensorflow[cpu-only]` 😞 

The main issue with just removing the dependency is that we actually do have minimum TensorFlow versions, and will probably bump them soon. Of course we can replace the requirement with a runtime check, but that has its own disadvantages.

For the time being, it should be possible to install scikeras with `pip install --no-deps scikeras`

stsievert:
> For the time being, it should be possible to install scikeras with pip install --no-deps scikeras

That might be worth adding to the docs.

adriangb:
Agreed, I'll add it in the install section.

adriangb:
@fcakyon I opened #222 to add some guidance in the docs. Do you feel that this would be an acceptable solution?

fcakyon:
@adriangb Unfortunately its not an acceptable solution in my case.

Im using this package in [this web app project](https://github.com/fcakyon/face-recognition-app-tutorial). Heroku creates the environment using [this requirements.txt](https://github.com/fcakyon/face-recognition-app-tutorial/blob/heroku/requirements.txt.

[...]

PR #222, opened 2021-03-22
DOC: Add note on installing without dependencies

adriangb:
Crossref #221

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -25,7 +25,7 @@ TensorFlow build or `tensorflow-cpu`_, you can instruct pip to ignore dependenci
     pip install --no-deps scikeras
 
 Then manually install compatible versions of TensorFlow and Scikit-Learn.
-As of SciKeras v0.2.1, the minimum required versions are as follows:
+As of SciKeras v0.3.3, the minimum required versions are as follows:
 
 - TensorFlow: v2.4.0
 - Scikit-Learn: v0.22.0
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -17,6 +17,19 @@ To install with pip, run:
 
 We recommend to use a virtual environment for this.
 
+To install without any dependencies, eg. to use a custom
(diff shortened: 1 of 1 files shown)


docs/source/install.rst
```diff
@@ -17,6 +17,19 @@ To install with pip, run:
...
 
+To install without any dependencies, eg. to use a custom
+TensorFlow build or `tensorflow-cpu`_, you can instruct pip to ignore dependencies:
+
+.. code:: bash
+
+    pip install --no-deps scikeras
+
+Then you can manually install any compatible TensorFlow and Scikit-Learn package.
+Currently, our minium versions are:
```
stsievert, line 28: This will go out of date unless a test is implemented or they're generated programmatically.

Maybe it's better to say "As of SciKeras v0.2.1, the minimum required versions are as follows:"

docs/source/install.rst
```diff
@@ -17,6 +17,19 @@ To install with pip, run:
...
 We recommend to use a virtual environment for this.
 
+To install without any dependencies, eg. to use a custom
+TensorFlow build or `tensorflow-cpu`_, you can instruct pip to ignore dependencies:
+
+.. code:: bash
+
+    pip install --no-deps scikeras
+
+Then you can manually install any compatible TensorFlow and Scikit-Learn package.
```
stsievert, line 27:
```suggestion
Then manually install compatible versions of TensorFlow and Scikit-Learn.
```

docs/source/install.rst
adriangb: Yeah that sounds good. I thought about something programmatic, but it gets complicated obviously. I think your text will do for now.

fcakyon:
@adriangb this pr doesnt solve the issue https://github.com/adriangb/scikeras/issues/221.
Heroku doesnt accept pip install `--no-deps`.

adriangb:
Ah yes, agreed. Sorry didn't mean to close #221. This is just extra documentation that can help in similar situations, but not the specific situation in #221, as you pointed out. I re-opened #221. I still don't have a solution for that.

PR #224, opened 2021-03-24
Unblock Python version

### LINKED ISSUE #223: Support to Python 3.9 ?
Hello,

thanks a lot for your work!

While installing scikeras I noticed that Python 3.9 it is currently not supported. I was wondering if you plan to support it in future.

Thanks in advance!

All the best,
Luca

adriangb:
Closes #223

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ version = "0.2.1"
 
 [tool.poetry.dependencies]
 importlib-metadata = {version = "^3.4.0", python = "<3.8"}
-python = ">=3.6.7"
+python = ">=3.6.7,<4"
 scikit-learn = "^0.22.0"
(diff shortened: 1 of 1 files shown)


stsievert:
Does this unpin the Python version as mentioned in https://github.com/adriangb/scikeras/issues/223#issuecomment-805945243? Why not totally remove the Python specification from SciKeras?

adriangb:
I don't think it can be removed altogether. It seems to be needed by Poetry. In fact, I tried just `>=3.6.7` but was forced to go with `>=3.6.7,<4`.

adriangb:
For the first question: this does not automatically make SciKeras compatible with 3.9, but the day that TensorFlow enables it we will be compatible as well.

PR #228, opened 2021-04-09
DOC: Add history attribute to migration

### LINKED ISSUE #227: History renamed compared to tf1 wrapper
I'm mostly through migration to this great project, as part of the move to upgrade to tf2 and noticed that `history` is called `history_` in scikeras. This is documented in the Model documentation but it might be nice to have a note in the migration guide.

As there are some PRs related to this history object I won't make a suggestion for the doc as I don't know how the behaviour changes (for my case it is just a case of renaming).

adriangb:
Closes #227

### DIFF
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -58,7 +58,7 @@ Calling ``fit`` resets this attribute, calling ``partial_fit`` on the other hand
    + losses = hist["mean_absolute_error"]
 
 .. note::
-   Unlike the TensorFlow wrappers, SciKeras normalizes the names of the keys, so that if you use `loss="mae"` you will get a key named `"mean_absolute_error"`.
+   Unlike the TensorFlow wrappers, SciKeras normalizes the names of the keys, so that if you use `metrics=["mae"]` you will get a key named `"mean_absolute_error"`.
 
 One-hot encoding of targets for categorical crossentropy losses
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -38,7 +38,7 @@ SciKeras does however have some backward incompatible changes:
 Fit returns ``self``
 ^^^^^^^^^^^^^^^^^^
 
-Previously calling ``fit`` on wrappers returned a Keras ``history`` dictionary.
+Previously calling ``fit`` on wrappers returned a Keras ``history`` object.
 To make wrappers more compatible with Scikit-Learn tools, ``fit`` now returns and instance of the estimator itself.
 Instead, the history is saved in the ``history_`` attribute.
 Calling ``fit`` resets this attribute, calling ``partial_fit`` on the other hand extends it.
--- a/docs/source/migration.rst
+++ b/docs/source/migration.rst
@@ -35,6 +35,21 @@ SciKeras is largely backwards compatible with the existing wrappers. For most ca
 
 SciKeras does however have some backward incompatible changes:
 
+Fit returns `self`
+^^^^^^^^^^^^^^^^^^
+
+Previously calling `fit` on wrappers returned a Keras `history` dictionary.
+To make wrappers more compatible with Scikit-Learn tools, `fit` now returns and instance of the estimator itself.
+Instead, the history is saved in the `history_` attribute.
+Calling `fit` resets this attribute, calling `partial_fit` on the other hand extends it.
+
+.. code:: diff
+
+   clf = KerasClassifier(...)
+   - hist = clf.fit(...)
+   + clf.fit(...)
+   + hist = clf.history_
+
 One-hot encoding of targets for categorical crossentropy losses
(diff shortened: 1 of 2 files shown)


docs/source/migration.rst
```diff
@@ -35,6 +35,20 @@ SciKeras is largely backwards compatible with the existing wrappers. For most ca
 
 SciKeras does however have some backward incompatible changes:
 
+Fit returns ``self``
+^^^^^^^^^^^^^^^^^^
+
+Previously calling ``fit`` on wrappers returned a Keras ``history`` object.
+To make wrappers more compatible with Scikit-Learn tools, ``fit`` now returns and instance of the estimator itself.
+Instead, the history is saved in the ``history_`` attribute.
+Calling ``fit`` resets this attribute, calling ``partial_fit`` on the other hand extends it.
```
stsievert, line 44: Hm... maybe this?

> In TF, calling ``Keras.fit`` returned a history object. However, in SciKeras, calling ``est.fit` returns the estimator to conform to the Scikit-learn API. The Keras history object is available through the `history_` attribute of the estimator:
>
> ``` diff
> def get_model():
>     ...
>     return model
>
> - model = get_model()
> - hist = keras.fit(model, ...)
> + clf = KerasClassifier(model=get_model, ...)
> + hist = clf.fit(...).history_
> ```

adriangb: Yeah that does seem a bit nicer, I'll incorporate it. Thank you.

stsievert: This example isn't 100% correct; see https://github.com/adriangb/scikeras/issues/227#issuecomment-816802724

adriangb: Yep you're right! The nuance of Keras returning a `History` object (which is a callback) that then contains the `history` attribute (which is a dict) was lost on me. Thank you for tracking it down and pointing it out!

stsievert: I think this warrants adding a `history` attribute to the wrapper. I think a warning should be added that `history` is an unstable attribute that can be deleted without warning.

adriangb: That only helps with the specific syntax of `clf.fit(...).history` but not with `clf.fit(...).{any other attribute or method}`. I think the previous behavior (not returning `self`) was problematic enough to warrant a breaking change. With this addition to the changelog we are properly documenting that change. So I plan to move forward with just this addition before releasing `v0.3.0`.

stsievert: This is my suggestion:

``` python
class BaseWrapper:
    ...

    @property
    def history(self):
        warnings.warn("this is an unstable feature and may be deleted without warning")
        return self.history_
```

I'm not sure how to interpret https://github.com/adriangb/scikeras/pull/228#discussion_r610918910

adriangb: What I am suggesting in https://github.com/adriangb/scikeras/pull/228#discussion_r610918910 is that:
1. The approach suggested in https://github.com/adriangb/scikeras/pull/228#discussion_r610922904 does not make us completely backwards compatible.
2. I think this change is warranted, and the fact that it is a breaking change is okay.
3. We are documenting the breaking change.

So I do not think there is a need for adding the `history` property and/or a warning. I think this PR will suffice.

stsievert: Why is the proposal in https://github.com/adriangb/scikeras/pull/228#discussion_r610922904 not a good (or at least a decent) idea?

adriangb: I have three main qualms with it:
1. In principle, I do not like replacing a complex object (`tf.keras.callbacks.History`) with another complex object (`self/BaseWrapper`) + a thin wrapper (https://github.com/adriangb/scikeras/pull/228#discussion_r610918910) and expecting it to be backwards compatible.
2. In practice, https://github.com/adriangb/scikeras/pull/228#discussion_r610922904 does not work if a user used to do `clf.fit(...).set_params` before since it would `tf.keras.callbacks.History.set_params` (this is documented method) but now would raise an `AttributeError`.
3. I think it is okay to be backwards incompatible for this change, especially if we document it.

The tl;dr is that instead of trying to be somewhat backwards compatible, but probably not being fully backwards compatible, I feel that it is better to just document the change and move on.

stsievert: How 'bout this implementation?

``` python
class BaseWrapper:
    ...

    @property
    def history(self):
        raise AttributeError(
            "class 'BaseWrapper' has no attribute 'history'.\n\n"
            "Similar attributes include an attribute named `history_`."
        )
```

adriangb: Would you want to put that in permanently, or say for 1 minor version?

I still think that a breaking change is okay, especially across packages, so we don't necessarily need to add this, the documentation should be enough.

stsievert: > Would you want to put that in permanently, or say for 1 minor version?

Doesn't matter. The same error type is raised, and the first sentence of the message is the same.

> I still think that a breaking change is okay, especially across packages, so we don't necessarily need to add this

Agreed. Documentation *should* be enough, but often is not enough (RTFD).

adriangb: Since the original issue was about documentation, and because I think the error is not strictly _necessary_, I would rather leave it out. Let's wait to see if the documentation is enough or if we get more issues before we add more code.  If we do get more related issues, we can totally add something like that at a later date. Sound good?

PR #233, opened 2021-06-01
ENH: Support routed parameters to callbacks

### LINKED ISSUE #232: Usage of callbacks__
Can you explain the usage of the routed parameter `callbacks__` from [Routed parameters](https://www.adriangb.com/scikeras/stable/advanced.html#routed-parameters) ?

- Is this a shortcut for the `fit__callbacks` ?
- Or can it be used to set parameters for each instantiated callback like the [example:  losses with routed parameters](https://www.adriangb.com/scikeras/stable/advanced.html#example-multiple-losses-with-routed-parameters) ?

I couldn't find a usage of the `route_params()` function with `destination="callbacks"`. Is this functionality already implemented ?

@stsievert

adriangb:
Closes #232 

TODO:
- [x] (maybe) support routing without the `fit__` prefix.

### DIFF
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -320,7 +320,6 @@ Additionally, SciKeras supports routed parameters to each individual loss, or to
         ...
         loss=[BinaryCrossentropy, CategoricalCrossentropy],
         loss__from_logits=True,  # BinaryCrossentropy(from_logits=True) & CategoricalCrossentropy(from_logits=True)
-        loss__label_smoothing=0.1,  # passed to each sub-item, i.e. `loss=[l(label_smoothing=0.1) for l in loss]`
         loss__1__label_smoothing=0.5,  # overrides the above, results in CategoricalCrossentropy(label_smoothing=0.5)
     )
     # or
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -391,7 +391,7 @@ Just like metrics and losses, callbacks support several syntaxes to compile them
     clf = KerasClassifier(
         ...
         callbacks=[tf.keras.callbacks.BaseLogger, tf.keras.callbacks.EarlyStopping]
-        callbacks__1__monitor="loss",
+        callbacks__1__monitor="loss",  # EarlyStopping(monitor="loss")
     )
     # or
     clf = KerasClassifier(
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -384,26 +384,29 @@ See the `Keras Metrics docs`_ for more details on mapping metrics to outputs.
 Callbacks
 +++++++++
 
-Just like metrics and losses, callbacks support several syntaxes to compile them depending on your needs:
-
 .. code:: python
 
     clf = KerasClassifier(
         ...
-        callbacks=[tf.keras.callbacks.BaseLogger, tf.keras.callbacks.EarlyStopping]
-        callbacks__1__monitor="loss",  # EarlyStopping(monitor="loss")
+        callbacks=tf.keras.callbacks.EarlyStopping
+        callbacks__monitor="loss",
     )
-    # or
+
+Just like metrics and losses, callbacks support several syntaxes to compile them depending on your needs:
+
+.. code:: python
+
+    # for multiple callbacks using dict syntax
     clf = KerasClassifier(
         ...
         callbacks={"bl": tf.keras.callbacks.BaseLogger, "es": tf.keras.callbacks.EarlyStopping}
         callbacks__es__monitor="loss",
     )
-    # or for a single callback
+    # or using list sytnax
     clf = KerasClassifier(
         ...
-        callbacks=tf.keras.callbacks.EarlyStopping
-        callbacks__monitor="loss",
+        callbacks=[tf.keras.callbacks.BaseLogger, tf.keras.callbacks.EarlyStopping]
+        callbacks__1__monitor="loss",  # EarlyStopping(monitor="loss")
     )
 
 Keras callbacks are event based, and are triggered depending on the methods they implement.
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -380,7 +380,9 @@ Here are several support use cases:
 See the `Keras Metrics docs`_ for more details on mapping metrics to outputs.
 
 Callbacks
-+++++++++
+^^^^^^^^^
+
+SciKeras can route parameters to callbacks.
 
 .. code:: python
 
@@ -439,7 +441,7 @@ simply use the ``fit__`` or ``predict__`` routing prefixes on your callback:
 Any routed constructor parameters must also use the corresponding prefix to get routed correctly.
 
 Routing as positional or keyword arguments
-++++++++++++++++++++++++++++++++++++++++++
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 It is possible that the consturctor of the class you need instantiated does not accept keyword arguments.
 In this case, instead of ``__name_of_kwarg=value`` you can use ``__0=value`` (or any other integer),
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -318,6 +318,16 @@ Example: callback with routed parameters
     clf.fit(X, y)  # stops after a couple iterations
 
 This allows you to apply hyperparameter tuning to your learning rate scheduler.
+SciKeras also supports passing a list of callbacks or even a dict of callbacks:
+
+.. code:: python
+
+    clf = KerasClassifier(
+        ...
+        callbacks=[tf.keras.callbacks.LearningRateScheduler, tf.keras.callbacks.EarlyStopping]
+        callbacks__0__0=Schedule,
+        callbacks__1__monitor="loss",
+    )
 
 Custom Scorers
 --------------
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -329,7 +329,7 @@ SciKeras also supports passing a list of callbacks or even a dict of callbacks:
         callbacks__1__monitor="loss",
     )
     # or
-        clf = KerasClassifier(
+    clf = KerasClassifier(
         ...
         callbacks={"lrs": tf.keras.callbacks.LearningRateScheduler, "es": tf.keras.callbacks.EarlyStopping}
         callbacks__lrs__0=Schedule,
--- a/docs/source/advanced.rst
+++ b/docs/source/advanced.rst
@@ -258,43 +258,187 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
 
 All routed parameters will be available for hyperparameter tuning.
 
+For example:
+
+.. code:: python
+
+   clf = KerasClassifier(..., fit__batch_size=32, predict__batch_size=10000)
+
 Below are some example use cases.
 
-Example: multiple losses with routed parameters
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Compilation with routed parameters
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+SciKeras can compile optimizers, losses and metrics with routed parameters.
+This allows hyperparameter tuning with deeply nested parameters to optimizer, losses and metrics.
+You can use this feature both when you let SciKeras compile your Model and when you compile your own model
+within ``model_build_fn`` (by accepting the ``compile_kwargs`` parameter).
+
+Optimizer
++++++++++
+
+.. code:: python
+
+    from scikeras.wrappers import KerasClassifier
+    from tensorflow import keras
+
+    clf = KerasClassifier(
+        model=model_build_fn,
+        optimizer=keras.optimizers.SGD,
+        optimizer__learning_rate=0.05
+    )
+
+
+Losses
+++++++
 
 .. code:: python
 
     from tensorflow.keras.losses import BinaryCrossentropy, CategoricalCrossentropy
 
     clf = KerasClassifier(
-        model=model_build_fn,
+        ...
+        loss=BinaryCrossentropy,
+        loss__label_smoothing=0.1,  # results in BinaryCrossentropy(label_smoothing=0.1)
+    )
+    # pure Keras object only syntax is still supported, but parameters can't be tuned
+    clf = KerasClassifier(
+        ...
+        loss=BinaryCrossentropy(label_smoothing=0.1),
+    )
+
+Keras, and SciKeras by extension, support passing a single loss (which will be applied to all outputs)
+or a loss for each output by passing a list of losses or a dict of losses.
+
+Additionally, SciKeras supports routed parameters to each individual loss, or to all losses together.
+
+.. code:: python
+
+    from tensorflow.keras.losses import BinaryCrossentropy, CategoricalCrossentropy
+
+    clf = KerasClassifier(
+        ...
         loss=[BinaryCrossentropy, CategoricalCrossentropy],
         loss__from_logits=True,  # BinaryCrossentropy(from_logits=True) & CategoricalCrossentropy(from_logits=True)
         loss__label_smoothing=0.1,  # passed to each sub-item, i.e. `loss=[l(label_smoothing=0.1) for l in loss]`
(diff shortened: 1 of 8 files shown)


tests/test_callbacks.py
```diff
@@ -36,9 +41,59 @@ def test_callbacks():
...
+    clf = KerasClassifier(
+        get_clf,
+        epochs=15,
+        loss="sparse_categorical_crossentropy",
+        optimizer=keras.optimizers.SGD(
+            learning_rate=0.1
+        ),  # to mirror https://keras.io/api/callbacks/learning_rate_scheduler/
+        fit__callbacks=[keras.callbacks.LearningRateScheduler],
+        fit__callbacks__0__0=Schedule,  # LRS does not acccept kwargs, only args, hence the 0__0 syntax
+        fit__callbacks__0__0__coef=0.2,  # translates to kwarg "coef" to the first arg of the first element of the callbacks kwarg to fit
```
stsievert, line 94: Suggested change:

``` python
kwargs = dict(model=get_clf, epochs=15, ...)
clf1 = KerasClassifier(**kwags,
        fit__callbacks={"lr": keras.callbacks.LearningRateScheduler},
        fit__callbacks__lr__0=Schedule,  # (LRS does not acccept kwargs)
        fit__callbacks__lr__0__coef=0.2,
)
clf2 = KerasClassifier(**kwags, 
        fit__callbacks=LearningRateScheduler(Schedule(coef=0.2)),
)
for clf in [clf1, clf2]:
    assert ...
```

tests/test_callbacks.py
```diff
@@ -36,9 +41,59 @@ def test_callbacks():
...
+
+    # using LRS with a nested instantiation of Schedule
+    clf = KerasClassifier(
+        get_clf,
+        epochs=15,
+        loss="sparse_categorical_crossentropy",
+        optimizer=keras.optimizers.SGD(
+            learning_rate=0.1
+        ),  # to mirror https://keras.io/api/callbacks/learning_rate_scheduler/
+        fit__callbacks=[keras.callbacks.LearningRateScheduler],
```
stsievert, line 92: Suggested change:

``` python
         fit__callbacks={"lr": keras.callbacks.LearningRateScheduler},
```

I think this will help with the ugly `0__0` syntax.

stsievert:
I am ambivalent on these changes being merged. The ugly syntax reduces my motivation. Is `fit__callbacks=LearningRateScheduler` possible?

tests/test_callbacks.py
adriangb: Would this be in-lieu of, or in addition to?

adriangb:
> I am ambivalent on these changes being merged. The ugly syntax reduces my motivation. Is `fit__callbacks=LearningRateScheduler` possible?

Thank you for the quick feedback! Is there anything else, aside from the ugly syntax, that makes you doubtful? Does it seem non-useful or anything?

I agree, it's not the prettiest. I think your proposal should work (it may work as-is, if not, I'm fairly certain it can be made to work).

tests/test_callbacks.py
adriangb: Incorporated as pytest parametrization

tests/test_callbacks.py
adriangb: Incorporated, keeping compatibility with the `0__0` syntax as well.

adriangb:
I was able to incorporate your feedback, and also support with/without the `fit__` prefix.

tests/test_callbacks.py
```diff
@@ -36,9 +44,76 @@ def test_callbacks():
...
+        dict(
+            callbacks=[keras.callbacks.LearningRateScheduler],
+            callbacks__0__0=Schedule,  # LRS does not acccept kwargs, only args, hence the 0__0 syntax
+            callbacks__0__0__coef=0.2,  # translates to kwarg "coef" to the first arg of the first element of the callbacks kwarg to fit
+        ),
+        dict(
+            callbacks={"lrs": keras.callbacks.LearningRateScheduler},
+            callbacks__lrs__0=Schedule,  # here the __0 is again because LRS does not accept kwargs
+            callbacks__lrs__0__coef=0.2,
+        ),
```
adriangb, line 84: @stsievert just want to check before I merge: does this look good to you? Thanks

stsievert:
Is it common to have only one callback? It might be worth adding support for not passing a list/dict to callbacks then. That'd enable this syntax:

``` python
KerasEstimator(...
    callbacks=keras.callbacks.LearningRateScheduler,
    callbacks__0=Schedule,
    callbacks__0__coef=0.2,
)
```

I think that's a lot cleaner. I suppose this could also be enabled by having a `callback` keyword arg instead of `callbacks`.

> The ugly syntax reduces my motivation. Is `fit__callbacks=LearningRateScheduler` possible?

I misspoke; I meant to say "is `fit__callbacks=LearningRateScheduler(Scheduler(coef=0.2))` possible? I think the docs should use that example if so.

> also support with/without the `fit__` prefix.

In Keras parlance, is "callback" synonymous with "fit callback"? Or, in SciKeras parlance, could I have `score__callbacks` and `predict__callbacks` in addition to `fit__callbacks`?

adriangb:
> Is it common to have only one callback?

I think it's fairly common.

>  I suppose this could also be enabled by having a callback keyword arg instead of callbacks

Keras does not have a `callback` arg, only a `callbacks` arg; I think we should follow that.
Sadly, the Keras docs are very lacking and only ever mention that the parameter is a `List of keras.callbacks.Callback instances`. But in practice it looks like using both a single instance and a dict are supported, so I guess that yes, we can support a single instance here as well, and I'll add a test for that. Thank you for the suggestion.

> In Keras parlance, is "callback" synonymous with "fit callback"? Or, in SciKeras parlance, could I have score__callbacks and predict__callbacks in addition to fit__callbacks?

This is actually a really good point. In Keras, you can have callbacks for both `fit` and `predict` (as kwargs).
Also, because of how [Keras'  event-based callback API](https://www.tensorflow.org/api_docs/python/tf/keras/callbacks/Callback#methods_2) works, a single callback can be "dispatched" for both fit and predict. This allows differentiating prediction during training (eg. for metric evaluation) and prediction during inference (when `predict` is called). I think we should support this is well, and we can do it within the Scikit-Learn API, without resorting to `**kwargs`. The only extra work we'll need to do is tie the `predict` callbacks initialization/destruction the the model's lifecycle (i.e. create them when `fit` or `initialize` are called and destroy/recreate them when either of those are called again, but not when `paritial_fit` is called).

docs/source/advanced.rst
```diff
@@ -258,43 +258,187 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+
+.. code:: python
+
+    from tensorflow.keras.losses import BinaryCrossentropy, CategoricalCrossentropy
+
+    clf = KerasClassifier(
+        ...
         loss=[BinaryCrossentropy, CategoricalCrossentropy],
         loss__from_logits=True,  # BinaryCrossentropy(from_logits=True) & CategoricalCrossentropy(from_logits=True)
         loss__label_smoothing=0.1,  # passed to each sub-item, i.e. `loss=[l(label_smoothing=0.1) for l in loss]`
```
stsievert, line 323:
```suggestion
```

I'm not seeing the need for this line; it's already expressed in the line above.

docs/source/advanced.rst
```diff
@@ -258,43 +258,187 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
         loss__1__label_smoothing=0.5,  # overrides the above, results in CategoricalCrossentropy(label_smoothing=0.5)
     )
+    # or
+    clf = KerasClassifier(
+        ...
+        loss={"out1": BinaryCrossentropy, "out2": CategoricalCrossentropy},
+        loss__from_logits=True,  # BinaryCrossentropy(from_logits=True) & CategoricalCrossentropy(from_logits=True)
+        loss__label_smoothing=0.1,  # passed to each sub-item, i.e. `loss={o: l(label_smoothing=0.1) for o, l in loss.items()}`
```
stsievert, line 331:
```suggestion
```

docs/source/advanced.rst
```diff
@@ -258,43 +258,187 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
 
-    from tensorflow.keras.losses import BinaryCrossentropy, CategoricalCrossentropy
+    from scikeras.wrappers import KerasClassifier
+    from tensorflow import keras
 
     clf = KerasClassifier(
         model=model_build_fn,
+        optimizer=keras.optimizers.SGD,
+        optimizer__learning_rate=0.05
+    )
```
stsievert, line 289:
```suggestion
    )
    clf = KerasClassifier(  # equivalent model
        model=model_build_fn, optimizer=keras.optimizers.SGD(learning_rate=0.5)
    )

.. note::

   Only routed parameters can be tuned; the object syntax can not be tuned.
   That is, in the example above, ``optimizer__learning_rate`` can be tuned with
   `sklearn.model_selection.RandomizedSearchCV`; `SGD(learning_rate=0.5)` can
   not be used with that class.
```

docs/source/advanced.rst
```diff
@@ -258,43 +258,187 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+        metrics={"out1": BinaryAccuracy, "out2": [BinaryAccuracy, AUC]},
+        metrics__out1__threashold=0.65,
+        metrics__out2__0__threashold=0.65,
+    )
+    # assuming you ordered your outputs like [out1, out2], this is equivalent to
+    clf = KerasClassifier(
+        ...
+        metrics=[BinaryAccuracy, [BinaryAccuracy, AUC]],
+        metrics__0__threashold=0.65,
+        metrics__1__0__threashold=0.65,
```
stsievert, line 370:
```suggestion
        metrics__1__0__threshold=0.65,  # AUC(threshold=0.65)
```

docs/source/advanced.rst
```diff
@@ -258,43 +258,187 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+        callbacks=[tf.keras.callbacks.BaseLogger, tf.keras.callbacks.EarlyStopping]
+        callbacks__1__monitor="loss",
+    )
+    # or
+    clf = KerasClassifier(
+        ...
+        callbacks={"bl": tf.keras.callbacks.BaseLogger, "es": tf.keras.callbacks.EarlyStopping}
+        callbacks__es__monitor="loss",
+    )
+    # or for a single callback
```
stsievert, line 394: This example should be moved up. It's probably the most common use case. I might specify the list/dict examples in a separate code block.

docs/source/advanced.rst
```diff
@@ -258,43 +258,187 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
++++++++++
+
+Just like metrics and losses, callbacks support several syntaxes to compile them depending on your needs:
+
+.. code:: python
+
+    clf = KerasClassifier(
+        ...
+        callbacks=[tf.keras.callbacks.BaseLogger, tf.keras.callbacks.EarlyStopping]
+        callbacks__1__monitor="loss",
```
stsievert, line 386:
```suggestion
        callbacks__1__monitor="loss",  # EarlyStopping(monitor="loss")
```

docs/source/advanced.rst
adriangb: It would actually be `metrics__1__0__threshold=0.65,  # BinaryAccuracy(threshold=0.65)`.

`metrics__1__0__threshold` translates to: `metrics[1][0](threashold=0.65)`

docs/source/advanced.rst
adriangb: Donzo, thanks for the suggestion

docs/source/advanced.rst
```diff
@@ -258,23 +258,217 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+    clf = KerasClassifier(
+        ...
+        metrics={"out1": BinaryAccuracy, "out2": [BinaryAccuracy, AUC]},
+        metrics__out1__threashold=0.65,
+        metrics__out2__0__threashold=0.65,
+    )
+    # assuming you ordered your outputs like [out1, out2], this is equivalent to
+    clf = KerasClassifier(
+        ...
+        metrics=[BinaryAccuracy, [BinaryAccuracy, AUC]],
```
stsievert, line 376:
```suggestion
        metrics=[BinaryAccuracy, [AUC, BinaryAccuracy]],
```

I think you should make sure AUC is selected. That way, when the reader tries to understand what object `metrics__1__0` selects, they can know if they're right or not.

adriangb: That sounds good.

docs/source/advanced.rst
```diff
@@ -258,23 +258,217 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+    )
+    clf = KerasClassifier(  # equivalent model
+        model=model_build_fn, optimizer=keras.optimizers.SGD(learning_rate=0.5)
+    )
+
+.. note::
+
+   Only routed parameters can be tuned; the object syntax can not be tuned.
+   That is, in the example above, ``optimizer__learning_rate`` can be tuned with
+   `sklearn.model_selection.RandomizedSearchCV`; `SGD(learning_rate=0.5)` can
```
stsievert, line 298: > That is, in the example above, it's possible to tune the learning rate with Scikit-Learn's `RandomizedSearchCV` by specifying `optimizer__learning_rate`. It's not possible to tune the learning rate with Scikit-Learn when `optimizer=SGD(learning_rate=0.5)` is specified.

There's also some double backticks missing (so text that should be monospaced will be italicized).

docs/source/advanced.rst
```diff
@@ -258,23 +258,217 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+
+    clf = KerasClassifier(
+        ...
+        loss=BinaryCrossentropy,
+        loss__label_smoothing=0.1,  # results in BinaryCrossentropy(label_smoothing=0.1)
+    )
+    # pure Keras object only syntax is still supported, but parameters can't be tuned
+    clf = KerasClassifier(
+        ...
+        loss=BinaryCrossentropy(label_smoothing=0.1),
```
stsievert, line 317: Suggestion:

``` python
clf = KerasClassifier(..., loss=BinaryCrossentropy(label_smoothing=0.1))
```

docs/source/advanced.rst
```diff
@@ -258,23 +258,217 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+Here are several support use cases:
+
+.. code:: python
+
+    from tensorflow.keras.metrics import BinaryAccuracy, AUC
+
+    clf = KerasClassifier(
+        ...
+        metrics=BinaryAccuracy,
+        metrics__threashold=0.65,
```
stsievert, line 358:
```suggestion
        metrics__threshold=0.65,
```

Also in several lines below.

docs/source/advanced.rst
```diff
@@ -258,23 +258,217 @@ of :py:class:`scikeras.wrappers.BaseWrappers`. Currently, they are:
...
+    clf = KerasClassifier(
+        ...
+        callbacks=tf.keras.callbacks.Callback,  # called from both fit and predict
+        fit__callbacks=tf.keras.callbacks.Callback,  # called only from fit
+        predict__callbacks=tf.keras.callbacks.Callback,  # called only from predict
+    )
+
+Any routed constructor parameters must also use the corresponding prefix to get routed correctly.
+
+Routing as args or kwargs
```
stsievert, line 443:
```suggestion
Routing as arguments/keyword arguments
```

"args" and "kwargs" are technical jargon.

docs/source/advanced.rst
adriangb: Great, thanks for the better text and catching the backticks.

docs/source/advanced.rst
adriangb: Thanks. My version was also missing a `,` (fixed it here and elsewhere).

docs/source/advanced.rst
adriangb: Thank you, incorporated.

docs/source/advanced.rst
adriangb: Good catch, fixed.

adriangb:
The nightly failures are the same in master, so I'm going to merge this regardless.

Issue #209, opened 2021-02-28
ENH: programmatic validations and error handling


adriangb:
From # : 

>  I think giving users better errors and validating their inputs like you are doing here can be a very valuable part of SciKeras, but currently it is done in an ad-hoc manner via `_check_model_compatibility`, etc. I think if we add more of these types of things, it would be nice to have an organized interface for it.

It would be good to organize these checks. We can split them into two categories:

- Checks before `fit` is called. This includes checking if the model is compiled, if it has a loss, that the number of outputs match the target, etc.
- Error handling after `fit` or `predict` are called: translating cryptic Keras/TF erorrs into user friendly errors with suggestions to fix them.

I envision something like this:

**wrappers.py**

```python3
class BaseWrapper(...):

    def __get_pre_fit_validators__(self) -> Generator[Callable[[BaseWrapper, Dict[str, Any]], None], None, None]:
        yield validate_compiled  # accepts self & self.model_.fit kwargs
        yield validate_has_loss
        yield validate_outptus_match_target
       ....

    def __handle_keras_exceptions__(self, exception: Exception, source: Literal["fit", "predict"]) -> Generator[Callable[[Exception], Literal["fit", "predict"]], None, None]:
        yield catch_some_cryptic_error  # can pass by just returning without raising
        yield catch_some_other_cryptic_error
        
```

Issue #111, opened 2020-10-18
ENH: move data validation to a modular interface


adriangb:
The main use case for this is to allow non-array inputs (perhaps dicts, datasets, etc.). The default implementation would remain the same, but users would have the ability to customize it for more advanced use cases.

Once #88 is merged (and after the next release), we coul it would move the internal data validation interface (`BaseWrapper._validate_data`) to a public transformer based interface, similar to the X/y processing. I would however keep them seperate for modularity. The main complication would be: do we have two transformers (one for X, one for y)? How can they coordinate (eg: to validate the same number of data points)?

An alternative would be to just take the interface we have and make it public.

In either case, the goal would be to have no private/non-customizable roadblocks to using arbitrary data inputs.

adriangb:
I think the easiest thing to do here may be to just make `validate_data` public, indicating that users can override it safely. That said, I'll probably wait until someone actually requests this as a feature before making any changes and expanding the public API.

adriangb:
Linking this to #166 since data validation is the main blocker for Datasets

PR #234, opened 2021-06-18
RFC: Composable input/output pipeline

adriangb:
This is an attempt to collect ideas from various issues / PRs into a coherent framework and generalize our current input/output transformers.

@stsievert  would you mind taking a look?

### DIFF
--- a/rfcs/1-input-pipeline.md
+++ b/rfcs/1-input-pipeline.md
@@ -1,3 +1,82 @@
+## Background
+
+One of the primary functions of the Scikit-Learn wrappers for Keras is fielding conversion from Scikit-Learn's data types
+to TensorFlow's data types. An example of these conversions is integer-encoding or one-hot encoding categorical targets.
+
+Originally, these conversions were hardcoded into the wrappers in an ad-hoc manner (see here).
+SciKeras introduced the concept of `TargetTransformer` and `InputTransformer`,
+two Scikit-Learn style transformers that formalize this data conversion framework and allow users to insert their own custom pipeline.
+
+Currently, user-customization requires subclassing the wrappers,
+and composability is only provided via meta-transformers (i.e. a Scikit-Learn pipeline) of transformers.
+
+Seperately, SciKeras also implements data validations that mirrors what most Scikit-Learn estimators implement,
+for example to assert that `X` and `y` are of the same length or that `y` is purely numeric for regressors.
+SciKeras also validates and inspects the model, for example to make sure that the output shape matches the
+target's shape.
+These checks are helpful for simple models, but may be too restrictive for more complex scenarios, like
+multi-input/output models.
+
+This RFC proposes a unifified interface for composable data transformations and validations.
+The goal is to provide a pipeline of default transformations and validations that cover the simple use cases,
+while allowing users to easily remove checks or add transformation steps for more advanced use cases.
+
+Some of the speicific functional requirements are:
+1. Able to implement the default (i.e. current) data transformations and validations. This includes:
+ - Integer encoding targets for classifiers.
+ - One-hot encoding targets for classifiers using the categorical crossentropy loss.
+ - Converting class probability predictions into class predictions for classifiers.
+2. Able to implement user-defined transformations, including:
+ - Splitting the input and/or target into multiple inputs/outputs.
+ - Reshaping 2D inputs into 3D.
+3. Able to operate on array-like data (lists, Numpy arrays, Pandas DataFrames, etc.) as well as `tf.data.Dataset`s.
+4. Composable and modifiable without subclassing.
+
+## Proposal
+
+This proposal consists of 2 pipelines:
+1. A pipeline for preparing array-like data for a `tf.data.Dataset`. This would include, for example, integer-encoding object-dtype arrays (`tf.Tensors` can't hold objects).
+2. A pipeline for applying transformations to `tf.data.Dataset`s. For example, one-hot encoding targets for classifiers using the categorical crossentropy loss.
+
+If the data comes in as a `tf.data.Dataset`, the first pipeline is skipped. If not, it is run and the output is converted to a `Dataset`.
+The second pipeline is then run in all cases.
+
+These pipelines will consist of chained transformers implementing a Scikit-Learn-like interface, but without being restricted to the exact Scikit-Learn API.
+
+```python
+from typing import Optional, Protocol, Sequence, Tuple, Type, Union
+
+from numpy import ndarray
+from numpy.typing import ArrayLike
+
+
+class ArrayTransformer:
+
+    def __init__(self, model: BaseWrapper) -> None:
+        self.model = model
+
+    def transform_input(self, X: ArrayLike, y: Optional[ArrayLike], sample_weight: Optional[ArrayLike], *, initialize: bool = True) -> Tuple[ndarray, Union[ndarray, None], Union[ndarray, None]]:
+        return X, y, sample_weight
+
+    def transform_output_proba(self, y_pred_proba: ndarray) -> ndarray:
+        return y_pred_proba
+
+
+class DatasetTransformer:
+
+    def __init__(self, model: BaseWrapper) -> None:
+        self.model = model
+
+    def transform_input(self, data: tf.data.Dataset, initialize: bool = True) -> tf.data.Dataset:
+        return data
+
+    def transform_output_proba(self, y_pred_proba: ndarray) -> ndarray:
+        return y_pred_proba
+```
+
+SciKeras will initialize the pipeline of transformers by passing a reference to the current estimator,
+and will then iterate through them as described above:
+
 ```python
 import itertools
 from typing import Optional, Protocol, Sequence, Tuple, Type, Union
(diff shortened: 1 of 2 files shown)


rfcs/1-input-pipeline.md
```diff
@@ -0,0 +1,391 @@
...
+
+class KerasRegressor:
+    def __init__(
+        self,
+        array_transformers = (RegressorInputValidator,)
+    ):
+        ...
+```
+
+## Issues this can potentially resolve
```
stsievert, line 377: These issues seem to be around modular validation/transforming. This RFC proposes one solution. What are other solutions? Why does this RFC represent the best solution?

rfcs/1-input-pipeline.md
```diff
@@ -0,0 +1,391 @@
+## Background
+
+One of the primary functions of the Scikit-Learn wrappers for Keras is fielding conversion from Scikit-Learn's data types
+to TensorFlow's data types. An example of these conversions is integer-encoding or one-hot encoding categorical targets.
+
+Originally, these conversions were hardcoded into the wrappers in an ad-hoc manner (see here).
```
stsievert, line 6: > ad-hoc manner (see here)

Link missing.

stsievert (commented):
Thanks for this RFC – this is easier to review than a PR. Here are some questions:

* What issue are you trying to solve with this RFC? What's an illustration of that issue?
* Is that issue user-facing?
* How does this RFC solve that (user-facing?) issue?

rfcs/1-input-pipeline.md
adriangb: Thank you, added. It was this: https://github.com/keras-team/keras/blob/80ca7a7a06058076fde5ee36f8ac1f2ca30f8996/keras/wrappers/scikit_learn.py#L155-L157

rfcs/1-input-pipeline.md
```diff
@@ -0,0 +1,391 @@
...
+## Proposal
+
+This proposal consists of 2 pipelines:
+1. A pipeline for preparing array-like data for a `tf.data.Dataset`. This would include, for example, integer-encoding object-dtype arrays (`tf.Tensors` can't hold objects).
+2. A pipeline for applying transformations to `tf.data.Dataset`s. For example, one-hot encoding targets for classifiers using the categorical crossentropy loss.
+
+If the data comes in as a `tf.data.Dataset`, the first pipeline is skipped. If not, it is run and the output is converted to a `Dataset`.
+The second pipeline is then run in all cases.
+
+These pipelines will consist of chained transformers implementing a Scikit-Learn-like interface, but without being restricted to the exact Scikit-Learn API.
```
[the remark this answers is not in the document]
adriangb: By "exact Scikit-Learn API" I mean roughly:
- having a `fit` method with the signature `fit(X, y=None) -> self`
- having a transform method with the signature `transform(X) -> X'`
- an `inverse_transform` method with the signature `inverse_transform(X') -> X`.
And no other methods (that are part of the API).
Also `inverse_transform(transform(X)) == X` is not a requirement of the API, but it certainly is the spirit of it (and how most sklearn transformers are implemented, when it makes sense to do so).

> And the proposed "Scikit-learn-like interface" is public, correct?

Yes, it would be public and consists of `ArrayTransformer` and `DatasetTransformer`.
These can be base classes, protocols/interfaces or just duck-typing (implementation detail).
The idea is that users could take our default data validation/transformations and mix/match with their own for use cases like multi-output models or multi-output class reweighting.

This proposed interface violates the sklearn interface in several ways:
- by passing `X`, `y` and `sample_weight` in `ArrayTransformer`
- by passing a `tf.data.Dataset` object in `DatasetTransformer`
- because these really aren't inverse transformations: the forward/input transformation transforms X & y, the output transformation transforms `y'` (y predictions) and/or `y` itself. So `inverse_transform(transform(X)) != X`.
- by allowing side effects to be applied to the `model` (this is an addition to the API, it doesn't strictly break the implementation, but it does break the spirit).

> How can this RFC be reworked to conform with the "exact Scikit-Learn API"?

I don't think it can be easily reworked: there are too many limitations in the sklearn transformer API ((and the authors have said as much themselves)[https://github.com/scikit-learn/scikit-learn/issues/3855#issuecomment-63213164]).

The good news is that the Scikit-Learn API is a strict subset of this API, so we could provide wrappers to convert the Scikit-Learn transformer API into this one (eg. by specifying what `X` is to your transformer and dispatching the methods).

But let me ask: why is it important to adhere to the Scikit-Learn API? I know there are good reasons, I just want understand which you are thinking of.
My thoughts are that if we can't interoperate directly (i.e. use `make_pipeline` and other facilities of sklearn), there is not much value in being semi-compatible. I also think there aren't that many things (beyond pipelines) that would be useful here. But this may be short-sighted, I'm open to other opinions.

adriangb:
> Thanks for this RFC – this is easier to review than a PR.

Good to know it helps! And thank you for reviewing.

> What issue are you trying to solve with this RFC? What's an illustration of that issue?

This solves some user-facing issues, as well as allowing us to rework things internally to be cleaner and more flexible for users.

There are several different issues, but I think the clearest one is #160.

The general user request is to be able to use `tf.data.Dataset`. 

We could hack this together as is (by hardcoding bypasses to `BaseWrapper._validate_data` and feature/target encoders), but even then we would lose attributes like `n_features_in_` that are both part of our API and the Scikit-Learn API. This RFC solves this problem by making processing of `tf.data.Dataset` a first-class citizen, and even allowing us to move some of our validation/preprocessing to use `tf.data.Dataset` which, for example, would allow us to provide `n_features_in_` and `classes_` regardless of if the input was array-like or a `tf.data.Dataset`.

The other issues linked are generally around the theme of more validations (#106, #143) or figuring out how to clean up our current validations (#111, #209). These two issues are intertwined because:
1. Every validation added has the chance of rubbing up against edge cases. If these validations aren't modular (eg if they're buried 3 layers deep into private method calls), we might block a use case we can't think of right now just to provide a nice error message.
2. If we add more ad-hoc private methods that do validation/transformation (eg `BaseWrapper._validate_data`, `BaseWrapper._check_model_compatibility`) we end up with a mess of private methods where it's hard to understand the execution order, goal of the method, etc.

This RFC resolves this by providing a unified interface for these validations/transformations that is modular, composable and public. I envision it helping our users (since they can disable or add validation/transformation steps) as well as us (the developers) because we can organize our default validation/transformations better.

rfcs/1-input-pipeline.md
adriangb: > What are other solutions?

This is, of course, a great question. I think general solutions would be variations of the same idea, perhaps less structured (eg. hardcoding that `tf.data.Dataset` inputs should skip `BaseWrapper._validate_data`). I'd have to think a bit to see if I can come up with any other structured approaches that might also solve the issues.

rfcs/1-input-pipeline.md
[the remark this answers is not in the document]
stsievert: This library has one job: to bring the Scikit-learn API to Keras. That means there needs to be a ***really good*** reason to break the Scikit-learn API.

That reason needs to consider alternative implementations, including the one that exactly follows the Scikit-learn API. I imagine some questions that need to be answered include the following:


* What affect does the implementation have on developers?
* What affect does the implementation have on the user?
* What specific issues make the implementation inappropriate?

rfcs/1-input-pipeline.md
```diff
@@ -0,0 +1,396 @@
...
+ArrayLike = Union[Sequence, np.ndarray]
+
+
+class NotInitializedError(Exception):
+    ...
+
+
+class ArrayTransformer:
+
+    def set_model(self, model: "BaseWrapper") -> None:
```
stsievert, line 66: :-1:

Why not follow the Scikit-Learn API?

``` python
class ArrayTransformer(BaseEstimator):
    def __init__(self, model=None):
        self.model = model

    ...

a = ArrayTransformer(model=model) # option 0

a = ArrayTransformer().model = model # option 1

a = ArrayTransformer()
a.set_params(model=model) # option 2
```

Yes, TF uses `set_model`. Why should their boilerplate be followed?

rfcs/1-input-pipeline.md
```diff
@@ -0,0 +1,396 @@
...
+        self.model = model
+
+    def transform_input(self, X: ArrayLike, y: Optional[ArrayLike], sample_weight: Optional[ArrayLike], *, initialize: bool = True) -> Data:
+        return X, y, sample_weight
+
+    def transform_output(self, y_pred_proba: np.ndarray, y: Union[np.ndarray, None]) -> Tuple[np.ndarray, Union[np.ndarray, None]]:
+        return y_pred_proba, None
+
+    def get_meta(self) -> Dict[str, Any]:
```
stsievert, line 75: :-1:

Why not have an attribute `meta_`?

rfcs/1-input-pipeline.md
```diff
@@ -0,0 +1,396 @@
...
+
+class DatasetTransformer:
+
+    def set_model(self, model: "BaseWrapper") -> None:
+        self.model = model
+
+    def transform_input(self, data: tf.data.Dataset, *, initialize: bool = True) -> tf.data.Dataset:
+        return data
+
+    def transform_output(self, y_pred_proba: np.ndarray, y: Union[np.ndarray, None]) -> Tuple[np.ndarray, Union[np.ndarray, None]]:
```
stsievert, line 87: Why not rename these functions `inverse_transform`?

rfcs/1-input-pipeline.md
stsievert: Why not have two Scikit-learn transformers, one for validation and one to change the data?

rfcs/1-input-pipeline.md
adriangb: The `get_meta` is stuck in there from older revision.

My current thoughts are to not have any interface for these meta parameters and instead let the implementers set attributes on `model` directly (or I guess we could have a `meta_` attribute on `BaseWrapper` and then do some `__getattr__` stuff to expose these attributes).

Sorry for having old stuff left in there.

rfcs/1-input-pipeline.md
adriangb: This would require users to give us a class to initialize (this object is being passed to the constructor that creates `model`).

I guess we could use parameter routing to allow setting any other parameters. Does this look any better?

```python
class MyTransf:

    def __init__(self, model, other):
        self.other = other
    
    ...

est = BaseWrapper(..., pipeline_param=[MyTransf], pipeline_param__0__other=True)
# or
est = BaseWrapper(..., pipeline_param={"tfname": MyTransf}, pipeline_param__ tfname__other=True)  # relies on dict ordering to know what order to run pipeline in
# or
est = BaseWrapper(..., pipeline_param=[("tfname", MyTransf)], pipeline_param__ tfname__other=True)  # same as an sklearn pipeline using a tuple to set the name
```

An alternative would be to require users to bind any other parameters using `functools.partial` or something.

rfcs/1-input-pipeline.md
adriangb: Because their signature and usage does not match `inverse_transform`'s signature. They wouldn't work in an sklearn `Pipeline`, nor would they be chainable with sklearn estimators.

rfcs/1-input-pipeline.md
adriangb: We would need:
1. Interface for validation using array-like data
2. Interface for data changing using array-like data
3. Interface for validation using Dataset
4. Interface for data changing using array-like data

I think this is too many interfaces. The only difference between a "validation" and a "transformation" is that the transformation _needs_ to return the data, but the validation does not. So by having the validation return the data we can collapse those two concepts into one. Validation transformers simply inspect the data but do not modify it.

stsievert: What use case/issue motivates having separate classes for Dataset/array-like inputs? Why not collapse it into one class?

Why is 4 interfaces "too many"? Why is the current framework with `target_encoder_` and `feature_encoder_` not sufficient?

I'm asking questions to get answers encoded in the RFC (an RFC is a really good idea to encode these decisions).

PR #240, opened 2021-08-11
Support sparse matrices for X

### LINKED ISSUE #239: Use of sparse matrices
Hi, nice to see someone continue development on a wrapper like this after the people at tensorflow decided to discontinue development on their wrapper.

I have run into an issue with the use of sparse matrices.

In the [API documentation](https://www.adriangb.com/scikeras/stable/generated/scikeras.wrappers.KerasClassifier.html#scikeras.wrappers.KerasClassifier) it is mentioned that the fit and predict functions from the KerasClassifier wrapper should work with array-like, sparse matrix and dataframe. However, when I use a sparse matrix, I get the following exception:

`TypeError: A sparse matrix was passed, but dense data is required. Use X.toarray() to convert to a dense numpy array.`

I used the quickstart guide to get a simple reproducable issue. I simply converted the ndarrays in the example into a scipy.sparse coo_matrix:

```
import numpy as np
from sklearn.datasets import make_classification
from tensorflow import keras
from scipy.sparse import coo_matrix

X, y = make_classification(1000, 20, n_informative=10, random_state=0)
X = X.astype(np.float32)
y = y.astype(np.int64)

X = coo_matrix(X)
y = coo_matrix(y)

def get_model(hidden_layer_dim, meta):
    # note that meta is a special argument that will be
    # handed a dict containing input metadata
    n_features_in_ = meta["n_features_in_"]
    X_shape_ = meta["X_shape_"]
    n_classes_ = meta["n_classes_"]

    model = keras.models.Sequential()
    model.add(keras.layers.Dense(n_features_in_, input_shape=X_shape_[1:]))
    model.add(keras.layers.Activation("relu"))
    model.add(keras.layers.Dense(hidden_layer_dim))
    model.add(keras.layers.Activation("relu"))
    model.add(keras.layers.Dense(n_classes_))
    model.add(keras.layers.Activation("softmax"))
    return model

clf = KerasClassifier(
    get_model,
    loss="sparse_categorical_crossentropy",
    hidden_layer_dim=100,
)

clf.fit(X, y)
y_proba = clf.predict_proba(X)

```
A potential reason for the issue could be that when validating the inputs via sklearn.utils.check_X_y, the default parameter for accept_sparse is False. [See also here](https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_X_y.html#sklearn.utils.check_X_y)

Setting this parameter to true might solve the issue (I will go and test that soon).
I am running this on python=3.7.10, scikit-learn=0.24.2 and tensorflow=2.5.0

adriangb:
Closes #239


adriangb:
Todo: update docs, maybe an example notebook?

docs/source/notebooks/sparse.md
```diff
@@ -0,0 +1,140 @@
...
+from sklearn.preprocessing import OneHotEncoder
+from sklearn.pipeline import Pipeline
+from tensorflow import keras
+```
+
+## Data
+
+The dataset we'll be using is designed to demostrate a worst-case/best-case scenario for dense and sparse input features respectively.
+It consists of a single categorical feature with equal number of categories as rows.
+This means the one-hot encoded representation will require as many columns as it does rows, making it very ineffienct to store as a dense matrix but very efficient to store as a sparse matrix.
```
carlogeertse, line 65: I think this is a perfect example, as data with many categorical features and/or categories per feature seems like the main reason to make use of a sparse matrix. Taking that to the extreme highlights the benefit of sparse matrices well.

docs/source/notebooks/sparse.md
```diff
@@ -0,0 +1,140 @@
...
+Our benchmark will be to just train each one of these pipelines and measure peak memory consumption.
+
+```python
+%memit dense_pipeline.fit(X, y)
+```
+
+```python
+%memit sparse_pipeline.fit(X, y)
+```
+
```
carlogeertse, line 121: Happy to see that you did know how to properly monitor memory usage.

docs/source/notebooks/sparse.md
```diff
@@ -0,0 +1,140 @@
...
+    [
+        ("encoder", OneHotEncoder(sparse=True, dtype=np.uint8)),
+        ("model", KerasRegressor(get_clf, loss="mse", epochs=5, verbose=False))
+    ]
+)
+```
+
+```python
+%memit sparse_pipline_uint8.fit(X, y)
+```
```
carlogeertse, line 140: 2 more things you might want to add:

- Monitor and mention the computation time for the sparse variant. While reduced memory usage is an obvious advantage of using sparse matrices, it came along with increased computation times (for my experiments at least). I'd be curious to see if you measure the same in these somewhat better setup experiments.
- Maybe mention the use of tf.data.Dataset along with a generator as an alternative solution to memory issues. I think this will work in most use cases. It just won't work when the scikeras wrapper is used inside _another_ wrapper or pipeline that doesn't support tf.data.Dataset. Which was the reason I needed to use a sparse matrix.

Not sure how relevant these 2 suggestions are, so I'll leave it up to you if you want include that information.

carlogeertse (approved):
Tutorial notebook looks good, explaining why and how to use sparse matrices. See my last comment for some minor suggestions.

docs/source/notebooks/sparse.md
adriangb: Great points. I added a run time measurement and mentioned Datasets, feel free to suggest changes to those sections

carlogeertse: Those additions look good to me.

mattalhonte-srm:
Heya! Just tested this and it doesn't work for me - converting it to lil makes the container blow up when I try to train, I need the csr matrix to stay a CSR matrix (TF can do much more efficient math on those!). The thing that worked for me was just passing `accept_sparse=True` and 0 other code changes.  Thanks!

adriangb:
Ouch. I forget why I had to put the conversion in there (I mean, there's a comment, but I'm sure some test failed).

PR #245, opened 2021-09-05
Fix handling of class_weight=balanced

adriangb:
Fixes use of `class_weight` parameter.

Currently, the `class_weight` parameter passed to the `KerasClassifier` constructor gets passed to Keras, which breaks `class_weight="balanced"` and any other values that are supported by Scikit-Learn (which we want to support in SciKeras) but no in Keras.

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -1357,7 +1357,7 @@ class KerasClassifier(BaseWrapper):
     @property
     def _fit_kwargs(self) -> Set[str]:
         # remove class_weight since KerasClassifier re-processes it into sample_weight
-        return BaseWrapper._fit_kwargs - set(("class_weight",))
+        return BaseWrapper._fit_kwargs - {"class_weight"}
 
     @staticmethod
     def scorer(y_true, y_pred, **kwargs) -> float:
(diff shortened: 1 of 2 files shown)


scikeras/wrappers.py
```diff
@@ -1354,6 +1354,11 @@ def _type_of_target(self, y: np.ndarray) -> str:
             target_type = type_of_target(self.classes_)
         return target_type
 
+    @property
+    def _fit_kwargs(self) -> Set[str]:
+        # remove class_weight since KerasClassifier re-processes it into sample_weight
+        return BaseWrapper._fit_kwargs - set(("class_weight",))
```
stsievert, line 1360:
```suggestion
        return BaseWrapper._fit_kwargs - {"class_weight"}
```

stsievert (commented):
Are any other values accepted besides `sample_weight="balanced"`?

adriangb:
> Are any other values accepted besides `sample_weight="balanced"`?

Nope, just `"balanced"` (and `None`): https://scikit-learn.org/stable/modules/generated/sklearn.utils.class_weight.compute_class_weight.html

PR #248, opened 2021-09-06
DRAFT: Avoid patching the Keras Model if Keras alrady supports __reduce__

adriangb:
This patching shouldn't be needed since https://github.com/keras-team/keras/pull/14748 was merged, pending https://github.com/keras-team/keras/issues/15300

### DIFF
--- a/scikeras/__init__.py
+++ b/scikeras/__init__.py
@@ -29,7 +29,8 @@ from tensorflow import keras as _keras
 from scikeras import _saving_utils
 
 
-_keras.Model.__reduce__ = _saving_utils.pack_keras_model
+if not hasattr(_keras.Model, "__reduce__"):
+    _keras.Model.__reduce__ = _saving_utils.pack_keras_model
 _keras.losses.Loss.__reduce__ = _saving_utils.pack_keras_loss
 _keras.metrics.Metric.__reduce__ = _saving_utils.pack_keras_metric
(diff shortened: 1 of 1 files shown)


stsievert:
Requiring a higher version of Keras/TF is not a unreasonable approach. Maybe that can be done at some point in the future.

adriangb:
yep, agreed, that may be a better alternative to this PR. I think next major release may be a good candidate.

adriangb:
Well I opened #252 to bump the versions.
Unfortunately, this is still not viable because we handle 2 things here which have still not been upstreamed:
- SavedModel in memory on Windows. I _think_ this will be fixed if https://github.com/keras-team/keras/pull/15551#pullrequestreview-801923609 gets merged. This is [here](https://github.com/adriangb/scikeras/blob/67de2388f34eecd1f09d4b1deaeb8ada5c60dc1b/scikeras/_saving_utils.py#L17-L23) in SciKeras.
- SavedModel does not correctly restore optimizer weights ([this code in SciKeras](https://github.com/adriangb/scikeras/blob/67de2388f34eecd1f09d4b1deaeb8ada5c60dc1b/scikeras/_saving_utils.py#L26-L51), [this bug report](https://github.com/keras-team/keras/issues/15512#issuecomment-964460397) in Keras).

So for now I am leaning towards bumping the versions in #252 and then looping back here once/if the above issues are fixed.

PR #252, opened 2021-11-09
BREAKING CHANGE: bump TensorFlow dependency to 2.7.0

adriangb:
This fixes compatibility with TensorFlow 2.7.0, at the cost of dropping support for older versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This package is available on PyPi:
 pip install scikeras
 ```
 
-The only dependencies are `scikit-learn>=0.22` and `TensorFlow>=2.4.0`.
+The only dependencies are `scikit-learn>=0.22` and `TensorFlow>=2.7.0`.
 
 You will need to manually install TensorFlow; due to TensorFlow's packaging it is not a direct dependency of SciKeras.
 You can do this by running:
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This package is available on PyPi:
 pip install scikeras
 ```
 
-SciKeras depends on `scikit-learn>=0.23.0` and `TensorFlow>=2.7.0`.
+SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
 
 You will need to manually install TensorFlow; due to TensorFlow's packaging it is not a direct dependency of SciKeras.
 You can do this by running:
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This package is available on PyPi:
 pip install scikeras
 ```
 
-SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
+The current version of SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
 
 You will need to manually install TensorFlow; due to TensorFlow's packaging it is not a direct dependency of SciKeras.
 You can do this by running:
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ This package is available on PyPi:
 pip install scikeras
 ```
 
-The only dependencies are `scikit-learn>=0.22` and `TensorFlow>=2.7.0`.
+SciKeras depends on `scikit-learn>=0.23.0` and `TensorFlow>=2.7.0`.
 
 You will need to manually install TensorFlow; due to TensorFlow's packaging it is not a direct dependency of SciKeras.
 You can do this by running:
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -32,7 +32,7 @@ You can also install SciKeras without any dependencies, for example to install a
 
     pip install --no-deps scikeras
 
-As of SciKeras v0.4.0, the minimum required versions are as follows:
+As of SciKeras v0.5.0, the minimum required versions are as follows:
 
 - TensorFlow: v2.7.0
 - Scikit-Learn: v1.0.0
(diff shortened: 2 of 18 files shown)


scikeras/_utils.py
```diff
@@ -1,60 +1,66 @@
...
+    orig_np_random_state = np.random.get_state()
+    if context.executing_eagerly():
+        tf_random_seed = context.global_seed()
+    else:
+        tf_random_seed = ops.get_default_graph().seed
 
-    def __enter__(self):
-        warnings.warn(
-            "Setting the random state for TF involves "
-            "irreversibly re-setting the random seed. "
```
stsievert, line 29: Why can this warning be removed?

scikeras/__init__.py
```diff
@@ -10,7 +10,7 @@
 __version__ = importlib_metadata.version("scikeras")
 
 
-MIN_TF_VERSION = "2.4.0"
+MIN_TF_VERSION = "2..7.0"
```
stsievert, line 13:
```suggestion
MIN_TF_VERSION = "2.7.0"
```

scikeras/_utils.py
```diff
@@ -1,60 +1,66 @@
...
 from tensorflow.keras import metrics as metrics_mod
 from tensorflow.keras import optimizers as optimizers_mod
+from tensorflow.python.eager import context
+from tensorflow.python.framework import config, ops
 
 
 DIGITS = frozenset(str(i) for i in range(10))
 
 
-class TFRandomState:
```
stsievert, line 21: Why rename `TFRandomState`?

README.md
```diff
@@ -22,7 +22,7 @@ This package is available on PyPi:
 pip install scikeras
 ```
 
-The only dependencies are `scikit-learn>=0.22` and `TensorFlow>=2.4.0`.
+SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
```
stsievert, line 25:
```suggestion
The current version of SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
```

docs/source/install.rst
```diff
@@ -34,8 +34,8 @@ You can also install SciKeras without any dependencies, for example to install a
 
 As of SciKeras v0.4.0, the minimum required versions are as follows:
```
stsievert, line 35:
```suggestion
As of SciKeras v0.5.0, the minimum required versions are as follows:
```

scikeras/_utils.py
adriangb: I believe the combination of seeing the global seed to None and config.disable_op_determinism() resolves the issue. But you are right to bring it up, I do want to double check

scikeras/_utils.py
adriangb: We don't need to rename it. I just did it while I was in the flow of reworking it to fix the actual issues. It is private, so in theory it doesn't hurt anything other than polluting this diff. I do like this style more, so I don't think it's worth it to revert just for the diff. Apologies for the confusion.

stsievert: If it's private, I would recommend removing it from the docs:

https://github.com/adriangb/scikeras/blob/67de2388f34eecd1f09d4b1deaeb8ada5c60dc1b/docs/source/notebooks/Benchmarks.md#L109

adriangb: I agree it should not be in the docs. But let's table that for a future discussion.

adriangb: Opened #253 to track

scikeras/_utils.py
adriangb: Here's my quick test:

```python
from typing import List
import tensorflow as tf

from scikeras._utils import tensorflow_random_state

without_seed: List[float] = []
with_seed: List[float] = []

for _ in range(10):
    # without setting the random state we should get distinct random values
    # and having set the random state previously does not carry over
    without_seed.append(tf.random.uniform([1]).numpy()[0])
    without_seed.append(tf.random.uniform([1]).numpy()[0])
    # if we set the random state twice, we get the same value
    with tensorflow_random_state(42):
        with_seed.append(tf.random.uniform([1]).numpy()[0])
    with tensorflow_random_state(42):
        with_seed.append(tf.random.uniform([1]).numpy()[0])

assert len(without_seed) == len(set(without_seed))
assert len(set(with_seed)) == 1
```

So, at least for this branch, setting the random state has no persistent global effect as far as I can tell.

stsievert: That quick test should go in the tests?

adriangb: Well, not directly: it is testing a private method and is also using TensorFlow directly, which is not reflective of this packages intent. But we could lift some of it (namely doing a couple loops to ensure reversibility of the global state) into the [existing higher level test](https://github.com/adriangb/scikeras/blob/67de2388f34eecd1f09d4b1deaeb8ada5c60dc1b/tests/test_parameters.py#L31) (and we probably want to remove [this test](https://github.com/adriangb/scikeras/blob/67de2388f34eecd1f09d4b1deaeb8ada5c60dc1b/tests/test_parameters.py#L65), which is just testing implementation details).

Does that sound reasonable to you?

stsievert: :+1: for not testing private functions.

Why shouldn't `tf_random_state` be public?

adriangb: No particular reason, just expanding the API. I'm more open to it now that it looks more reliable + it's simplified into to a function based CM instead of a class.

stsievert: Let's move this conversation to #253.

PR #262, opened 2022-01-05
MAINT: add optional packages for tensorflow and tensorflow-cpu; update documentation

### LINKED ISSUE #261: tensorflow should be a main dependency (instead of a dev dependency)
I am trying to use `scikeras` in a poetry based project. I ran

```
poetry add scikeras
```
which worked. At runtime, I got an error saying that tensorflow 2.7.0 was needed. This should've been caught at the time I added the package during poetry. The only reason, from my understanding that it wasn't was because tensorflow is a dev dependency, rather than a real dependency, for poetry.

metasyn:
# Description

Fixes #261 

# Changes

```
poetry add --optional tensorflow==2.7.0
poetry add --optional tensorflow==2.7.0
```

Update documentation. I updated all pip install sites - but maybe only some need to be updated?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,13 +19,15 @@ An overview of the advantages and differences as compared to the TF wrappers can
 This package is available on PyPi:
 
 ```bash
-pip install scikeras
+# Normal tensorflow
+pip install scikeras[tensorflow]
+
+# or tensorflow-cpu
+pip install scikeras[tensorflow-cpu]
 ```
 
 The current version of SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
 
-You will need to manually install TensorFlow; due to TensorFlow's packaging it is not a direct dependency of SciKeras.
-You can do this by running:
 
 ```bash
 pip install tensorflow
--- a/README.md
+++ b/README.md
@@ -26,12 +26,17 @@ pip install scikeras[tensorflow]
 pip install scikeras[tensorflow-cpu]
 ```
 
+SciKeras packages TensorFlow as an optional dependency because there are
+several flavors of TensorFlow available (`tensorflow`, `tensorflow-cpu`, etc.).
+Depending on _one_ of them in particular disallows the usage of the other, which is why
+they need to be optional.
+
+`pip install scikeras[tensorflow]` is basically equivalent to `pip install scikeras tensorflow`
+and is offered just for convenience. You can also install just SciKeras with
+`pip install scikera`s, but you will need a version of tensorflow installed at
+runtime or SciKeras will throw an error when you try to import it.
+
 The current version of SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
-
-
-```bash
-pip install tensorflow
-```
 
 ### Migrating from `tf.keras.wrappers.scikit_learn`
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -19,13 +19,15 @@ An overview of the advantages and differences as compared to the TF wrappers can
...
+pip install scikeras[tensorflow]
+
+# or tensorflow-cpu
+pip install scikeras[tensorflow-cpu]
 ```
 
 The current version of SciKeras depends on `scikit-learn>=1.0.0` and `TensorFlow>=2.7.0`.
 
-You will need to manually install TensorFlow; due to TensorFlow's packaging it is not a direct dependency of SciKeras.
-You can do this by running:
```
adriangb, line 28: I think it's worth leaving a comment in explaining why we recommend installing with extras. Something like:

> SciKeras packages TensorFlow as an optional dependency because there are several flavors of TensorFlow available (`tensorflow`, `tensorflow-cpu`, etc.). `pip install scikeras[tensorflow]` is equivalent to `pip install scikeras tensorflow` and is offered just for convenience. You can also install just SciKeras with `pip install scikeras`, but you will need a version of tensorflow installed at runtime or SciKeras will throw an error when you try to import it.

adriangb (changes requested):
Looks good, only issue is the pinned versions

adriangb:
Thinking about this a bit more: do you see any advantage in making these optional dependencies `scikeras[tensorflow]` v.s. just editing the docs to read `pip install scikeras tensorflow` or `pip install scikeras tensorflow-cpu` @metasyn ? I'm open to either and I think both need the same explanation blurb in the docs, I'm just trying to think what might be less confusing for a user.

metasyn:
> Thinking about this a bit more: do you see any advantage in making these optional dependencies `scikeras[tensorflow]` v.s. just editing the docs to read `pip install scikeras tensorflow` or `pip install scikeras tensorflow-cpu` @metasyn ? I'm open to either and I think both need the same explanation blurb in the docs, I'm just trying to think what might be less confusing for a user.

I guess adding it as an optional dependency makes it more obvious, in my opinion, in terms of what is going on. It reminds me of using `coverage[toml]` or `dask[complete]` - where its expected that w/o that, certain things wont work. It would be really nice if poetry somehow supported an "either or" kind of constraint or something.

I'm happy to change whatever you prefer though.

adriangb:
I think this change works. I agree with your assessment of it being more "obvious" even if it's just thin syntactic sugar.

> It would be really nice if poetry somehow supported an "either or" kind of constraint or something.

It's not necessarily a poetry issue (in fact, poetry kinda support this usage when installing a local package via dependency groups). Really this is a pip / package ecosystem issue: https://github.com/pypa/setuptools/issues/1139

pyproject.toml
```diff
@@ -33,10 +33,15 @@ importlib-metadata = {version = "^3", python = "<3.8"}
 python = ">=3.7.0,<3.10.0"
 scikit-learn = ">=1.0.0"
 packaging = ">=0.21,<22.0"
-tensorflow = {version = "2.7.0", optional = true}
-tensorflow-cpu = {version = "2.7.0", optional = true}
+tensorflow = {version = ">=2.7.0", optional = true}
+tensorflow-cpu = {version = ">=2.7.0", optional = true}
+
+[tool.poetry.extras]
```
metasyn, line 39: Ahh, thanksx, I didn't realize I had to add this too.

adriangb:
Thanks @metasyn !

PR #267, opened 2022-02-26
Add sklearn mixin inheritance

go95:
Sklearn interface presumes the estimators inherit from RegressorMixin and ClassifierMixin. It doesn't accually add any useful functionality, but it allows one to check, what interface do the class implement

`isinstance(model, RegressorMixin)`

It is useful to check this in your code to know if e.g. the class implements predict_proba and how do predict work (give some continuous scores or discrete predictions).

### DIFF
--- a/scikeras/wrappers.py
+++ b/scikeras/wrappers.py
@@ -9,7 +9,7 @@ from typing import Any, Callable, Dict, Iterable, List, Mapping, Set, Tuple, Typ
 import numpy as np
 import tensorflow as tf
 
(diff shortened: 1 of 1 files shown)


adriangb:
Would you mind running the pre-commit hooks please?

go95:
Thanks, @adrian. I've complied lint now. Not sure, how one would trigger a precommit-hook. Could you trigger it once again?

adriangb (approved):

PR #278, opened 2022-07-07
chore/fix: compatibility with newest nightly builds

adriangb:
- Downgrades an overly aggressive/ambitious numeric comparison
- Removes samples with sample weight = 0


adriangb:
Spoke with the scikit-learn team w.r.t lowering precision of their estimator checks so that we don't have to do juggling like we did here with sample weights. They're working on it, but nothing concrete so far. For now, I think our best option is to move forward with this. I will think about making a PR to their repo and then clean stuff up here (the whole separation into high and low precision falls into that category as well)

PR #279, opened 2022-07-20
Allow functools.partial in wrappers.py

jpgard:
Allow for model instantiation functions which are built using functools.partial.

I imagine this is a common use case, where model functions might need to have certain parameters set using functools.partial. In that case, inspect.function() returns False, and this extra check is required.

Thanks for the awesome package, super useful!!!


adriangb:
Seems reasonable. Docs failure is unrelated. Can you just add a small test?

PR #285, opened 2022-10-08
Fix sample_weight performance problem

richmanbtc:
## Problem

- The performance is degraded due to the wrong way of giving sample_weight to keras api.
- https://github.com/keras-team/keras/pull/16177#discussion_r820162317

## Solution

- Fixed to give sample_weight to keras api in the correct way.


adriangb:
So this "fix" makes legitimate tests fail. I'd also like to see a notebook showing performance degradation and how this fix addresses it.

PR #308, opened 2023-10-05
chore: updating the poetry version to fix the push issue

cristian-rincon:
This change includes the update on the pyproject.toml file. As the previous release failed due to the package version, as it was the same one that was already present in the Pypi. 

Please refer to this traceback to see the whole error: https://github.com/adriangb/scikeras/actions/runs/6383670304

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -112,7 +112,7 @@ jobs:
     strategy:
       matrix:
         tf-version: [2.11.0]
-        python-version: ["3.7", "3.9"]
+        python-version: ["3.8", "3.9"]
         sklearn-version: [1.0.0]
 
     steps:
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -34,7 +34,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.8", "3.9", "3.10", "3.11"]
-        tensorflow-version: ["2.12.0"]
+        tensorflow-version: ["2.12.0", "2.13.0"]
       fail-fast: false
 
     steps:
@@ -111,7 +111,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [2.12.0]
+        tf-version: [2.12.0, 2.13.0]
         python-version: ["3.8", "3.9"]
         sklearn-version: [1.0.0]
 
@@ -150,7 +150,7 @@ jobs:
       matrix:
         os: [MacOS, Windows]  # test all OSs except Ubuntu, which is already running other tests
         python-version: ["3.8", "3.11"]  # test only the two extremes of supported Python versions
-        tensorflow-version: ["2.12.0"]  # test only the two extremes of supported TF versions
+        tensorflow-version: ["2.12.0", "2.13.0"]  # test only the two extremes of supported TF versions
       fail-fast: false
 
     steps:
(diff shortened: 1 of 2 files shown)


adriangb:
So I think the problem here is that removing the pin on grpcio allowed poetry to resolve a newer tensorflow version (see https://github.com/adriangb/scikeras/actions/runs/6424229105/job/17444399467?pr=308#step:6:164, not sure what it was before) and TensorFlow minor updates usually (and unfortunately) have breaking changes. So we'll need the PR to bump the minium TensorFlow version to 2.13 or so and fix any issues that causes. Would you be willing to work on that @cristian-rincon ?

cristian-rincon:
Deprecating tf 2.11 solved several pipelines. I will take a look to see if the current code supports 2.13

cristian-rincon:
Hi @adriangb Upgrading to TensorFlow 13 right now will require several refactors. I cannot afford those refactors now. However, after conducting several tests with different constraints, I have found a solution that works and can fix the pipelines. I suggest we release this version with the new constraints for now. We can then address the upgrade in future releases. What do you think?

pyproject.toml
```diff
@@ -27,15 +27,16 @@ license = "MIT"
...
+version = "0.11.1"
 
 [tool.poetry.dependencies]
 importlib-metadata = {version = ">=3", python = "<3.8"}
 python = ">=3.8.0,<3.12.0"
 scikit-learn = ">=1.0.0"
 packaging = ">=0.21"
-tensorflow = {version = ">=2.11.0", optional = true}
-tensorflow-cpu = {version = ">=2.11.0", optional = true}
+tensorflow = {version = "<=12.0.0", optional = true}
```
adriangb, line 37: Is this version a typo? Did you mean `2.12.0`?

cristian-rincon: You're right! i will apply the refactor now

cristian-rincon: Done! ✅ @adriangb

pyproject.toml
```diff
@@ -27,15 +27,16 @@ license = "MIT"
...
 
 [tool.poetry.dependencies]
 importlib-metadata = {version = ">=3", python = "<3.8"}
 python = ">=3.8.0,<3.12.0"
 scikit-learn = ">=1.0.0"
 packaging = ">=0.21"
-tensorflow = {version = ">=2.11.0", optional = true}
-tensorflow-cpu = {version = ">=2.11.0", optional = true}
+tensorflow = {version = "<=2.12.1", optional = true}
+tensorflow-cpu = {version = "<=2.12.1", optional = true}
```
adriangb, line 38: Why `2.12.1`? Can we be compatible with any `2.12.X` version at least?

cristian-rincon: Done ✅

cristian-rincon: @adriangb

pyproject.toml
```diff
@@ -27,15 +27,16 @@ license = "MIT"
 name = "scikeras"
 readme = "README.md"
 repository = "https://github.com/adriangb/scikeras"
-version = "0.11.0"
+version = "0.11.1"
```
adriangb, line 30: Lets make this a minor version bump.

.github/workflows/tests.yaml
```diff
@@ -111,8 +111,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        tf-version: [2.11.0]
-        python-version: ["3.7", "3.9"]
+        tf-version: [2.12.0]
+        python-version: ["3.8", "3.9"]
```
adriangb, line 115: Does 2.12 not support 3.7?

cristian-rincon: Hi, unfortunately no.

adriangb (approved):

PR #330, opened 2024-11-03
poetry: update tensorflow-io-gcs-filesystem version to <=0.37.1

clstaudt:
Required to run `poetry install` successfully on macOS & Python 3.11


adriangb:
Lots of errors 😢

clstaudt:
How to proceed? An update of the dependency was what allowed me to run poetry install in the first place.

PR #334, opened 2024-12-12
Mark project as deprecated

adriangb:


adriangb:
I think the easiest thing here will be to just skip tests for this release. As long as there are no code changes I think that's safe. I'll revert the changes to dependencies.

The main point is to put the warning on import, in the docs and in the README.

adrinjalali:
It's there now 🎉

https://keras.io/api/utils/sklearn_wrappers/

fiendish:
Checking on this

